PR #33, opened 2025-07-27
feat: New CLI auth for Omnara runs

ishaansehgal99:
Omnara login and omnara connect


ishaansehgal99:
addressed already in separate PR

Issue #82, opened 2025-08-13
[BUG] Terminal Emulator Copy/Paste issues


ksarangmath:
**Describe the bug**
When using Warp, pasting into the terminal causes the entire omnara session to hang

**To Reproduce**
Steps to reproduce:
1. Use warp and create an omnara session
2. Paste ~50 lines of text

**Expected behavior**
Pasting should not cause the entire session to hang

**Environment:**
- Python version: 3.12
- Omnara version: 1.4.9

wolfgangmeyers:
I noticed in the claude code session that there was a "Pasted N+ lines" in the input. I hit enter and it continued. Seems like the issue here is that the large prompts end up getting "pasted" instead of submitted as a prompt.

cpursley:
Can confirm - also using Warp. Everything completely freezes when copying and pasting.

wolfgangmeyers:
> Can confirm - also using Warp. Everything completely freezes when copying and pasting.

I'm a newcomer - what is Warp?

cpursley:
> I'm a newcomer - what is Warp?

https://www.warp.dev

ishaansehgal99:
thanks for sharing guys, yes we are aware this is a warp issue! normal terminal app works. I think it has to do with how buffer is handled - will investigate

ishaansehgal99:
This issue has been resolved! It should work now - making a new release

ishaansehgal99:
Awesome this is fixed in release 1.14.15

PR #102, opened 2025-08-14
fix: Create Copy/Paste STDIN Buffer for PTY

ishaansehgal99:
We needed to buffer the data ourselves when the PTY buffer is full, rather than trying to force it through or discarding it. This creates a smooth data flow pipeline: `Warp → stdin → our buffer → PTY → Claude`


ishaansehgal99:
Resolving https://github.com/omnara-ai/omnara/issues/82

PR #106, opened 2025-08-14
fix: Handle dots in Claude project path conversion

n1330:
Claude Code project directories appear to convert dots to hyphens, similar to how slashes are converted, as shown in the following example:

Claude Code execution directory:
/Users/foo/src/github.com/omnara-ai/omnara

Claude Code Project directory:
/Users/foo/.claude/projects/-Users-foo-src-github-com-omnara-ai-omnara

Omnara's existing code did not handle dot conversion, which prevented Claude Code responses from being retrieved when running in such directories. 
This commit fixes the issue.


ksarangmath:
Thanks for the PR!

PR #152, opened 2025-08-20
Amp streaming progress

wolfgangmeyers:
These updates allow amp to stream messages incrementally to the UI as tool calls are made.


ksarangmath:
Just tried it out again! There's some edge cases to be fixed, but it's almost there. Similar spot to where I was with the Claude Code wrapper for a while. I'm able to see the tool calls, but it stops streaming the chat when I switch from terminal to mobile, and back to terminal. Maybe a one-off thing.

I noticed you have the AMP is processing your request... message appear in the chat, is that so the user can get an indication that AMP is thinking? I was thinking of adding some loading animation instead in the UI to indicate that, what do you think of that, and then removing the AMP is processing your request... message?

ksarangmath:
Thanks for putting this PR up @wolfgangmeyers ! I'll have to close it since I'm cleaning up the repo for a new version of Omnara, but I hope you'll contribute to the new version!

PR #208, opened 2025-09-06
feat: Session Monitoring

ishaansehgal99:
Detect session status using hearbeating


backend/api/agents.py
```diff
@@ -302,6 +302,11 @@ def notification_callback(connection, pid, channel, payload):
 
                         yield f"event: git_diff_update\ndata: {json.dumps(data)}\n\n"
 
+                    elif event_type == "presence_update":
```
ksarangmath: nit: could we name the event type something like "heartbeat", just so an llm can easily grep it

servers/fastapi_server/routers.py
```diff
@@ -208,6 +210,69 @@ def update_title_with_session():
...
+        instance.last_heartbeat_at = datetime.now(timezone.utc)
+        db.commit()
+        db.refresh(instance)
+
+        # Send NOTIFY to the per-instance channel so SSE clients can react immediately
+        try:
+            channel_name = f"message_channel_{agent_instance_id}"
+            payload = json.dumps(
+                {
+                    "event_type": "presence_update",
```
ksarangmath: same as above

ksarangmath (approved):
🚀🚀🚀

PR #255, opened 2025-09-29
add gemini integration

rbagga:
still a bit buggy but basic functionality is there


rbagga:
gemini integration is more robust now. we now fork the gemini-cli source code as a submodule and inject / extract messages like we do with codex.  The gemini fork is here: https://github.com/rbagga/gemini-cli

ksarangmath:
Thanks for putting this PR up, really appreciate it! Unfortunately, I'll have to close it since we're moving to a new version of Omnara, but I hope you'll contribute to the new version!

PR #257, opened 2025-10-01
n8n node default agent id and type

asehgal4:
- Updated the n8n node so it generates a default agent id and type specific to each workflow execution
- Also allows the user to enter a non specific agent id rather than a generated UUID
- The generated $$agent\\_id ≈ hash(workflow\\_id \mathbin\Vert execution\\_id \mathbin\Vert agent\\_type, BASE)$$
- Note: Subworkflows will have distinctly generated agent ids as they have different workflow ids


src/integrations/n8n/src/nodes/Omnara/Omnara.node.ts
```diff
@@ -1,19 +1,20 @@
...
-	INodeType,
-	INodeTypeDescription,
-	NodeOperationError,
-	NodeConnectionType,
-	SEND_AND_WAIT_OPERATION,
+    IExecuteFunctions,
+    INodeExecutionData,
+    INodeType,
+    INodeTypeDescription,
+    NodeConnectionTypes,
```
ksarangmath: I don't think this exists

src/integrations/n8n/src/nodes/Omnara/Omnara.node.ts
```diff
@@ -30,8 +31,8 @@ export class Omnara implements INodeType {
 		defaults: {
 			name: 'Omnara',
 		},
-		inputs: [NodeConnectionType.Main],
-		outputs: [NodeConnectionType.Main],
+        inputs: [NodeConnectionTypes.Main],
```
ksarangmath: same here, can revert these lines

src/integrations/n8n/src/utils/GenericFunctions.ts
```diff
@@ -57,3 +58,36 @@ export function formatMessageResponse(message: any): IDataObject {
...
+
+export function getAgentType(this: IExecuteFunctions, index: number): string {
+	const userProvidedAgentType = this.getNodeParameter('agentType', index) as string;
+	if (userProvidedAgentType && userProvidedAgentType.trim() !== '') {
+		return userProvidedAgentType;
+	}
+	
+	// Use workflow name as default
+	const workflowName = this.getWorkflow().name as string;
+	return workflowName || 'Omnara Agent';
```
ksarangmath: nit:
```suggestion
	return workflowName || 'n8n Agent';
```

src/integrations/n8n/README.md
```diff
@@ -99,19 +99,21 @@ The webhook also provides:
 
 For workflows triggered by other events (cron, manual, other apps).
 
-**Requirements:**
-1. Generate a unique UUID for Agent Instance ID (use Expression: `{{ $uuid() }}`)
+**Requirements (for custom agent ids and types):**
+1. Generate or define an Agent Instance ID
```
ksarangmath: can we change these instructions, since we don't have to generate/define an agent instance id, or choose agent type, etc? maybe you can just say you can use the nodes out of the box, and manually define the message or let AI agent generate the message.

ksarangmath (changes requested):
looks good, main thing is that you're importing NodeConnectionTypes, which errors

src/integrations/n8n/src/nodes/Omnara/Omnara.node.ts
asehgal4: I think it's a package discrepancy issue from n8n. NodeConnectionTypes exists for me

ksarangmath: You're right

ksarangmath (approved):

PR #261, opened 2025-10-08
fix: restore --name parameter

EduardoWeber:
## Summary

- Fixed the `--name` parameter that was being ignored by the Claude wrapper
- Modified `claude_wrapper_v3.py` to read `OMNARA_AGENT_DISPLAY_NAME` environment variable

## Behavior

**Before this fix:**
Running `omnara --name AgentName` would ignore the passed name and use the Claude Code default name.

**After this fix:**
Running `omnara --name AgentName` correctly uses "AgentName" as the agent display name.

## Details

The CLI was already setting the `OMNARA_AGENT_DISPLAY_NAME` environment variable, but the wrapper wasn't reading it. This change makes the wrapper check the environment variable first, falling back to the default name if not set.


ishaansehgal99:
LGTM!

PR #302, opened 2026-08-05
fix: Allow creating machine daemon tokens with PAT or org tokens

cmsparks:
There was a bug where we restricted creation of daemon tokens. This fixes that (and removes some attribution columns.


asherdale (approved):
lgtm

PR #303, opened 2026-08-05
feat: Add grant editing functionality

cmsparks:
Let's add grant editing for:
* machine grants
* model grants


cmsparks:
2 comments from prev version: 

```
should we show/hide this action based on org-manage access? ProjectGrantsPage admits project admins through project.access.can_manage_access, but both useMachinePool and UpdateProjectMachinePoolGrant require org.manage
```

Resolving this UX issue


```
is it intentional for these edits to affect existing agents? this validates the grant only against the configured model’s current revision, but Resolver.Resolve reapplies the live grant to the revision pinned by an existing model context

for example, a 96k context-window grant valid against the current 128k revision would make an existing context pinned to an older revision with 64k fail even though the API says updates apply only to new launches
```

I think it should affect existing agents, because that's how it works with env var overrides with machine pools. though obviously the documentation seems inaccurate. I'll fix this

cmsparks:
Bypassing merge approval because was approved previously

PR #305, opened 2026-08-05
chore: Openapi spec cleanup

cmsparks:
* Improve documentation
* Remove unnecessary openapi spec docs
* strictly type metadata
* strictly marshal delta frames
* fix visible project and machine types


internal/httpapi/response_contract_integration_test.go
```diff
@@ -0,0 +1,171 @@
...
+	buffering bool
+	hijacked  bool
+	body      bytes.Buffer
+}
+
+func (r *responseContractRecorder) WriteHeader(status int) {
+	if !r.decided {
+		r.decided = true
+		r.status = status
+		mediaType, _, err := mime.ParseMediaType(r.Header().Get("Content-Type"))
```
ksarangmath: Could we decide whether to buffer and validate based on the response declared by the matched OpenAPI operation, rather than the response’s actual Content-Type? If a JSON endpoint accidentally omits or changes its content type, buffering becomes false and we skip both body and content-type validation.

api/openapi/openapi.yaml
```diff
@@ -7256,13 +7253,27 @@ components:
       additionalProperties: true
       x-go-type: json.RawMessage
       x-go-type-skip-optional-pointer: true
+    Metadata:
+      type: object
+      description: Arbitrary key-value metadata. Maximum 16 pairs, keys up to 64 characters, values must be strings of up to 512 characters.
+      x-go-type: json.RawMessage
```
ksarangmath: Could this generate a typed map[string]string instead of json.RawMessage? The schema says metadata has string values and shared limits, but json.RawMessage lets internal code store any JSON and return responses that violate that schema. For example, default machine-pool metadata is still map[string]any, and the storage layer only checks that it is an object.

I think metadata should have a shared Go type and validator, with conversion to raw JSON happening only when reading from or writing to JSONB. If secrets or runtime metadata intentionally have different rules, they should use separate types instead.

api/openapi/openapi.yaml
```diff
@@ -7256,13 +7253,27 @@ components:
...
+      x-go-type-skip-optional-pointer: true
+      maxProperties: 16
+      additionalProperties:
+        type: string
+        maxLength: 512
```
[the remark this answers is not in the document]
ksarangmath: Also MCPOAuthStartRequest can accept 16 user entries, then mcpOAuthSecretMetadata always adds mcp_url, so the server stores 17 entries and can return a Secret that violates this same schema. Side note, should mcp_url be a typed field rather than put in metadata? And more generally, this is a smell pointing at whether the metadata should be a user defined thing that is just returned to the user, vs something we add to for our app logic. If both are use cases, we probably don't want a single metadata type dictating both of those cases, probably need more domain specific definitions.

api/openapi/openapi.yaml
```diff
@@ -9187,14 +9186,40 @@ components:
...
     VisibleMachine:
-      allOf:
-        - $ref: '#/components/schemas/MachineSummary'
-        - type: object
-          additionalProperties: false
-          required: [access]
-          properties:
-            access:
-              $ref: '#/components/schemas/MachineAccess'
+      type: object
```
ksarangmath: This is a bigger change, but could we upgrade the using OpenAPI 3.1.2 and model this through a shared field component rather than copying all of MachineSummary here? Currently we're using OpenAPI 3, which is already pretty old so we should probably upgrade (oapi-codegen didn't have support for 3.1 when we started out), and the changes in this PR would be made much cleaner if we had OpenAPI 3.1

I don’t think MachineSummary itself should be the extensible base because it is also a concrete, closed response type. The clean shape seems to be an open MachineSummaryFields component, with both MachineSummary and VisibleMachine composing those fields and closing their final shape with unevaluatedProperties: false. We could do the same with ProjectFields, Project, and VisibleProject.

That would leave one owner for the shared fields and prevent the concrete schemas from drifting. oapi-codegen v2.8 can generate the complete API with this shape.

internal/httpapi/response_contract_integration_test.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	contract, err := loadResponseContract()
+	if err != nil {
+		panic(fmt.Sprintf("load openapi response contract: %v", err))
+	}
+	route, pathParams, err := contract.router.FindRoute(request)
+	if err != nil {
+		return
+	}
+	if route.Operation.Responses.Status(recorder.status) == nil {
```
ksarangmath: Could we express this fallback in the OpenAPI contract itself using reusable 4XX and 5XX responses whose body is Error, and remove the special undeclared-error path here?

Exact status definitions take precedence over range responses, so we can keep known errors such as 400, 401, and 404 documented precisely while still covering errors produced by shared middleware. Then every accepted response is represented in the contract, and any status without an exact or range response is a real violation.

internal/httpapi/response_contract_integration_test.go
```diff
@@ -0,0 +1,171 @@
...
+			Request:    request,
+			PathParams: pathParams,
+			Route:      route,
+		},
+		Status:  recorder.status,
+		Header:  recorder.Header(),
+		Options: &openapi3filter.Options{IncludeResponseStatus: true},
+	}
+	input.SetBodyBytes(recorder.body.Bytes())
+	if err := openapi3filter.ValidateResponse(context.Background(), input); err != nil {
```
ksarangmath, line 372: Once the spec moves to 3.1.2, could we make the JSON-body part of this check document-aware rather than relying on Kin (Kin doesn't properly validate certain things from json 2020-12)?

Kin’s 3.1 path uses santhosh-tekuri/jsonschema/v6 internally, but it compiles the selected response schema separately from the root OpenAPI document. A composed response containing a reference such as `#/components/schemas/MachineSummaryFields` can therefore fail compilation, after which Kin silently falls back to its legacy validator. That validator does not enforce `unevaluatedProperties`, so a VisibleMachine or VisibleProject response with an unexpected field can still pass this check.

Since we already use santhosh-tekuri/jsonschema/v6, we can register the complete OpenAPI document and validate against that. Kin can remain responsible for the route, status, media type, and headers. The document-aware pass would validate only the outgoing JSON body in integration tests.

We could also add a regression that proves that a response with shared fields passes while the same response with an unexpected property fails. If Kin eventually passes on that, we can remove the extra body validation here.

internal/httpapi/response_contract_integration_test.go
```diff
@@ -0,0 +1,171 @@
...
+func (r *responseContractRecorder) Hijack() (net.Conn, *bufio.ReadWriter, error) {
+	hijacker, ok := r.ResponseWriter.(http.Hijacker)
+	if !ok {
+		return nil, nil, errors.New("response writer does not support hijacking")
+	}
+	r.hijacked = true
+	return hijacker.Hijack()
+}
+
+// validateResponseContract panics on contract violations so every integration
```
ksarangmath: Could we make this harness fail closed for every /api/v1 response it wraps? `!recorder.decided` represents an implicit empty 200 in net/http, not the absence of a response, and a `FindRoute` failure means the implementation has no matching OpenAPI operation. Silently returning in either case breaks the core invariant that every integration response is checked against the contract. Any intentional exclusions, such as hijacked or streaming responses, should be explicit.

internal/model/stream.go
```diff
@@ -11,6 +16,87 @@ type StreamEvent struct {
...
+		return nil, fmt.Errorf("stream event block index %d is negative", e.BlockIndex)
+	}
+	switch e.Kind {
+	case StreamEventBlockStart:
+		if e.Block == nil {
+			return nil, errors.New("block_start stream event requires a block")
+		}
+		return json.Marshal(struct {
+			Kind       StreamEventKind `json:"kind"`
+			BlockIndex int             `json:"block_index"`
```
ksarangmath: Since every block-scoped event now always serializes block_index, could we also make it required in the corresponding OpenAPI variants and remove the default?

internal/model/stream.go
```diff
@@ -11,6 +16,87 @@ type StreamEvent struct {
 	Error      *StreamError    `json:"error,omitempty"`
 }
 
+// MarshalJSON emits the public ModelOutputStreamDelta wire shape: exactly the
```
ksarangmath: Could we keep StreamEvent as the internal normalized model event and introduce a separate public API delta type at the stream boundary? Adding MarshalJSON here makes the internal execution type own the OpenAPI wire representation, which couples internal/provider evolution to public API evolution. The stream sink already performs the public translation so that seems like the clearer boundary for constructing the wire type.

api/openapi/openapi.yaml
```diff
@@ -7753,12 +7769,16 @@ components:
...
+          description: Monotonic frame sequence within this model call context's preview stream, starting at 1. Delta frames carry no SSE id; gaps in seq mean dropped best-effort frames.
         source_seq_start:
           $ref: '#/components/schemas/AgentSequence'
+          description: First provider stream event coalesced into this frame.
         source_seq_end:
           $ref: '#/components/schemas/AgentSequence'
+          description: Last provider stream event coalesced into this frame.
         coalesced_count:
           $ref: '#/components/schemas/AgentCount'
+          description: Number of provider stream events coalesced into this frame.
```
ksarangmath: nit: Could we make coalesced_count part of the completeness contract here? A frame is intact only when coalesced_count == source_seq_end - source_seq_start + 1. The sink can coalesce across a source-sequence gap, but the frontend currently checks only for gaps between frames, so it misses loss hidden inside one frame. Since the sink always emits this field, could we make it required and have the frontend reject frames where the count does not match the range?

ksarangmath (commented):
Two main things:
1 - Some metadata go types have different validation than what's in openapi, and that could be because semantically all metadata doesn't mean the same thing (maybe some are external user inputted blobs vs internal application logic things). We could decide that all metadata is public facing and adhere's to the metadata type you made in openapi, but we should make that explicit in the go validation too
2 - based on the openapi changes, we should probably migrate to openapi 3.1

api/openapi/openapi.yaml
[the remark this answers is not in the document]
cmsparks: The intent of metadata is so that users can filter on their own custom metadata, and then our own implementation can also filter on metadata (sort of like the metadata we described with )

MCP URL is just used for filtering in the frontend, so I don't think it's much of a smell to be attached as metadata, though we should fix the start_request validation regardless, maybe have it be a stricter subtype of the metadata type.

FWIW wrt the cursor comment, I think it's fine to break existing metadata at the moment to just make our metadata storage more consistent (though in the future we can't really do that)

api/openapi/openapi.yaml
cmsparks: Added this

cmsparks:
One note: technically this has the breaking change of if there's large metadata blocks that are in the DB previously, this might break parsing of some rows. This shouldn't cause problems in our current DB.

The only case is the machine platform object, which is fixed via a migration + converting to a single string.

api/openapi/openapi.yaml
```diff
@@ -8961,8 +9514,7 @@ components:
           minimum: 1
           maximum: 2147483647
         metadata:
-          type: object
-          additionalProperties: true
+          $ref: '#/components/schemas/Metadata'
```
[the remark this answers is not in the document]
cmsparks: comment here: https://github.com/omnara-ai/omnara/pull/305#issuecomment-5247166410

internal/resourcemeta/resourcemeta.go
```diff
@@ -0,0 +1,59 @@
...
+		m = Metadata{}
+	}
+	raw, err := json.Marshal(m)
+	if err != nil {
+		return nil, fmt.Errorf("encode metadata: %w", err)
+	}
+	return raw, nil
+}
+
+func FromJSON(raw json.RawMessage) (Metadata, error) {
```
ksarangmath, line 69: for both JSON and FromJSON, should we validate the metadata as well? like checking the same constraints that are in the openapi spec?

internal/storage/queries/machine_daemon.sql
```diff
@@ -161,8 +161,8 @@ SET last_observed_at = statement_timestamp(),
     provider_runtime_mismatch_since = NULL,
     wake_attempt_expires_at = NULL,
     metadata = CASE
-      WHEN sqlc.arg(observed_platform)::jsonb = '{}'::jsonb THEN metadata
-      ELSE jsonb_set(metadata, '{observed_platform}', sqlc.arg(observed_platform)::jsonb, true)
+      WHEN sqlc.arg(observed_platform)::text = '' THEN metadata
+      ELSE jsonb_set(metadata, '{observed_platform}', to_jsonb(sqlc.arg(observed_platform)::text), true)
```
ksarangmath, line 165: I think there might be a bug where this could set a metadata dict with 16 values to 17. Could we either make these columns, or reserve the field similar to mcp_url, or something liek that

ksarangmath (approved):
LGTM!

internal/httpapi/response_contract_integration_test.go
```diff
@@ -0,0 +1,363 @@
...
+		}
+		return
+	}
+	if route.Operation.Responses.Status(recorder.status) == nil {
+		panic(fmt.Sprintf(
+			"openapi response contract violation: %s %s returned undeclared status %d\nresponse body: %s",
+			request.Method, request.URL.Path, recorder.status, recorder.body.Bytes(),
+		))
+	}
+	if !recorder.buffering {
```
ksarangmath, line 346: very nit: should we still check Content-Type against openapi for non buffered responses

internal/storage/queries/machine_daemon.sql
cmsparks: good point added a shared function + type to validate this

PR #310, opened 2026-08-06
fix: Implement fix for machine env overlay on machine bindings

cmsparks:
Setting env on a machine binding (i.e. machine in an agent config) does not get propagated to the startup script or daemon on machine start. This fixes that bug.


api/openapi/openapi.yaml
```diff
@@ -7244,7 +7244,7 @@ components:
         cwd:
           type: string
         env_overlay:
-          description: Env overlay applied to processes launched through this binding; it does not change the machine environment applied to the provider machine at provisioning. A null entry removes the key.
+          description: Env overlay applied to processes launched through this binding. For pool bindings it is also applied to the machine environment when the pool machine is provisioned, so the provider_options startup_script and the daemon run with it; connected machines keep their own environment. A null entry removes the key.
```
asherdale, line 7291: should we update the description for secret_env_overlay too?

asherdale (commented):
lgtm - one note though is that `make verify` fails for this right now, might want to check that out

api/openapi/openapi.yaml
cmsparks: Yeah, i'll update this and fix make verify

cmsparks: Done — secret_env_overlay now carries the same pool-binding provisioning note.

asherdale (approved):

PR #312, opened 2026-08-06
feat: reconcile default resources across organizations

asherdale:
- add `omnara-api reconcile-defaults plan|apply` so operators can preview or apply configured defaults across existing organizations
- update cluster-managed machine pool defaults while preserving every organization's existing capacity limits and leaving existing machines unchanged
- reconcile default model provider settings and configured models while retaining provider credentials, explicit project opt-outs, and customized grants

### DIFF
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"io/fs"
 	"log/slog"
 	"net/http"
@@ -30,10 +31,15 @@ import (
 	"github.com/omnara-ai/omnara/internal/secrets"
 	"github.com/omnara-ai/omnara/internal/storage"
 	"github.com/omnara-ai/omnara/internal/storage/identitystore"
+	"github.com/omnara-ai/omnara/internal/storage/orglifecycle"
 	"github.com/omnara-ai/omnara/internal/toolcatalog"
 )
 
-const outboundHTTPClientTimeout = 30 * time.Second
+const (
+	outboundHTTPClientTimeout  = 30 * time.Second
+	defaultReconciliationPlan  = "plan"
+	defaultReconciliationApply = "apply"
+)
 
 func main() {
 	cfg, err := config.Load()
@@ -41,6 +47,14 @@ func main() {
 		log := slog.New(logpkg.NewJSONHandler(os.Stdout, nil))
 		log.Error("load config", "error", err)
 		os.Exit(1)
+	}
+	if len(os.Args) > 1 {
+		if err := runDefaultReconciliation(context.Background(), cfg, os.Args[1:], os.Stdout); err != nil {
+			log := slog.New(logpkg.NewJSONHandler(os.Stdout, nil))
+			log.Error("run command", "error", err)
+			os.Exit(1)
+		}
+		return
 	}
 	if err := cfg.ValidateAPI(); err != nil {
 		log := slog.New(logpkg.NewJSONHandler(os.Stdout, nil))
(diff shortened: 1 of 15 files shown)


internal/storage/modelstore/default_reconciliation.go
```diff
@@ -0,0 +1,290 @@
...
+			))
+			if apply {
+				if _, err := updateConfiguredModelTx(
+					ctx,
+					qtx,
+					configuredModelUpdateFromDefault(model.ID, desired),
+					management.Cluster,
+				); err != nil {
+					return nil, nil, fmt.Errorf("update default configured model %q: %w", modelTemplate.Name, err)
+				}
```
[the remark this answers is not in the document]
asherdale: this is intentional. project model grants are live, tenant-owned policy, so defaults reconciliation preserves them rather than silently rewriting their overrides. if a new configured-model revision makes an override incompatible, subsequent model resolution fails closed with `invalid_model_policy`; the project admin can then lower or clear that override

asherdale:
CI passed here: https://github.com/omnara-ai/omnara/actions/runs/31130942725

internal/storage/modelstore/default_reconciliation.go
```diff
@@ -0,0 +1,290 @@
...
+		}
+		project, err := qtx.GetProjectByIdempotencyKey(
+			ctx,
+			dbsqlc.GetProjectByIdempotencyKeyParams{
+				OrgID: current.OrgID, IdempotencyKey: defaultProjectIdempotencyKey,
+			},
+		)
+		if errors.Is(err, pgx.ErrNoRows) {
+			warnings = append(warnings, fmt.Sprintf(
+				"org %s: skip configured models because the default project is missing",
```
ksarangmath: could we still reconcile configured models even if the default project is missing?

cmd/api/main.go
```diff
@@ -208,6 +222,65 @@ func main() {
 	}
 }
 
+func parseDefaultReconciliationMode(args []string) (string, error) {
```
ksarangmath, line 225: I remember you said that this reconcile function should be in this service, but I forget the reason. Would it be cleaner if we made a new binary/script to run the reconciler? Seems kinda off putting it in this API service

internal/storage/orglifecycle/default_reconciliation.go
```diff
@@ -0,0 +1,59 @@
...
+
+func (s *Service) ReconcileDefaults(
+	ctx context.Context,
+	input ReconcileDefaultsInput,
+) (ReconcileDefaultsResult, error) {
+	txOptions := pgx.TxOptions{}
+	if !input.Apply {
+		txOptions.AccessMode = pgx.ReadOnly
+	}
+	tx, err := s.pool.BeginTx(ctx, txOptions)
```
ksarangmath, line 109: nit: looks like we do the whole reconciliation in one transaction, would a for loop be better in case one org doesnt work it doesn't stop every org from updating, or just in general we seem to lock a lot of things in one big transaction which seems like it might not scale well

internal/storage/modelstore/default_reconciliation.go
```diff
@@ -0,0 +1,290 @@
...
+	"errors"
+	"fmt"
+	"math"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/omnara-ai/omnara/internal/storage/internal/dbsqlc"
+	"github.com/omnara-ai/omnara/internal/storage/management"
+)
+
+const defaultProjectIdempotencyKey = "default"
```
ksarangmath: nit: could this const be shared with wherever this is actually defined when we create the default project

internal/storage/executionstore/default_reconciliation.go
```diff
@@ -0,0 +1,78 @@
...
+func (s *Store) ReconcileDefaultMachinePoolsTx(
+	ctx context.Context,
+	tx pgx.Tx,
+	templates []DefaultMachinePoolTemplate,
+	apply bool,
+) ([]string, error) {
+	qtx := s.q.WithTx(tx)
+	var changes []string
+	for _, template := range templates {
+		if err := ValidateDefaultMachinePoolTemplate(template); err != nil {
```
ksarangmath: nit: Could this pass reconciliation with a pool template that would fail api startup? We validate after replacing the template limits with each org’s existing limits

ksarangmath (approved):
LGTM, probably will need some edits though to handle the new runtime protection field on machine pools, and the min fields that you're adding

cmd/api/main.go
asherdale: as discussed, leaving this in the API service to avoid additional deployment surface

internal/storage/modelstore/default_reconciliation.go
asherdale: good call, fixed this

internal/storage/orglifecycle/default_reconciliation.go
asherdale: agreed, changed this to one transaction per org. a failing org rolls back independently and reconciliation keeps going for the rest, and retries are idempotent

internal/storage/executionstore/default_reconciliation.go
asherdale: fixed this by running the same shared template/provider validation used at API startup before applying each org's preserved limits, then validating the effective per-org config too

PR #313, opened 2026-08-06
fix: Improve keying for block_id to mitigate risk of Slack failing to submit actions.

cmsparks:


cmsparks:
Closing bc i couldn't repro, probably worth it to keep an eye out on this

PR #317, opened 2026-08-07
feat: support live MCP config changes

asherdale:
- Allows live agent config updates to add, change, or remove MCP declarations instead of rejecting the request.
- Reconciles MCP connection rows at model-work boundaries, reusing unchanged connections, reinitializing changed servers, and expiring removed servers without a steady-state lock transaction.
- Keeps pending tool calls pinned to their originating MCP config, then applies endpoint, authentication, permission, and tool-surface changes on the next model round.

### DIFF
--- a/internal/storage/executionstore/agent_mcp_connections_store.go
+++ b/internal/storage/executionstore/agent_mcp_connections_store.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/jackc/pgx/v5"
+	"github.com/omnara-ai/omnara/internal/agentconfig"
 	"github.com/omnara-ai/omnara/internal/storage/internal/dbsqlc"
 	"github.com/omnara-ai/omnara/internal/storage/storeerr"
 )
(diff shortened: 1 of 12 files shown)


internal/storage/executionstore/agents_integration_test.go
```diff
@@ -1249,22 +1241,15 @@ mcp:
...
-	events, err := store.Execution().ListAgentEventsForRead(ctx, testProjectID, launch.Agent.ID, 0, 100)
+	loaded, err := store.Execution().GetAgentInProject(ctx, testProjectID, launch.Agent.ID)
 	if err != nil {
-		t.Fatalf("list events after rejected change: %v", err)
+		t.Fatalf("load agent after change: %v", err)
 	}
-	if len(events) != len(initialEvents) {
-		t.Fatalf("rejected change appended events: before=%d after=%d events=%+v", len(initialEvents), len(events), events)
+	if loaded.CurrentConfigID != changed.AgentConfig.ID {
+		t.Fatalf("current config = %s, want %s", loaded.CurrentConfigID, changed.AgentConfig.ID)
```
cmsparks, line 1252: This test just checks that the config changed, can we make it a bit more precise? 

and/or would be nice to add an e2e test which tests the whole flow in service_e2e_mcp_test.go

internal/storage/executionstore/agent_mcp_connections_store.go
```diff
@@ -40,58 +41,89 @@ type MCPConnectionRecord struct {
...
-		dbsqlc.LockAgentInProjectParams{ProjectID: input.ProjectID, ID: input.AgentID},
+		dbsqlc.LockAgentInProjectParams{ProjectID: projectID, ID: agentID},
 	); err != nil {
 		if errors.Is(err, pgx.ErrNoRows) {
-			return MCPConnectionRecord{}, storeerr.ErrNotFound
+			return nil, storeerr.ErrNotFound
 		}
-		return MCPConnectionRecord{}, fmt.Errorf("lock agent for mcp connection: %w", err)
-	}
-	row, err := qtx.GetOrCreateMCPConnection(ctx, dbsqlc.GetOrCreateMCPConnectionParams{
```
cmsparks, line 78: nit: This structure feels like a code smell, I think there's probably a better way to structure the stuff after checking for reconciled, my read is that the flow is as follows:

```
list servers
check if reconciled
createAgentMCPConnectionsTx from all servers (which loops through and does one query per server)
list servers again
expire connections afterwards (loops through and does one query per server)
```

Can we combine the 2nd list operation and server expiration into just one query? We can probably have it as just a single query instead of two query loops, something like this:


```
-- name: ExpireRemovedMCPConnections :exec
UPDATE agent_mcp_connections c
SET state = 'expired', mcp_session_id = '', tools_snapshot = '[]'::jsonb,
    generation = c.generation + 1, updated_at = statement_timestamp()
FROM agents agent
WHERE agent.project_id = sqlc.arg(project_id)
  AND agent.id = sqlc.arg(agent_id)
  AND c.agent_id = agent.id
  AND c.server_key <> ALL(sqlc.arg(server_keys)::text[])
  AND c.state <> 'expired';
```

cmsparks (approved):

internal/storage/executionstore/agents_integration_test.go
asherdale: tightened the storage test to assert the active runtime contract has the expected `docs` MCP server, and added a thorough service e2e test

internal/storage/executionstore/agent_mcp_connections_store.go
asherdale: good call, replaced the second list + per-connection expiration loop with a single `ExpireRemovedMCPConnections` query after the upserts. it preserves the existing cleanup/generation behavior and handles an empty server list too

PR #320, opened 2026-08-07
feat: add content block presentation metadata

asherdale:
- add opaque metadata to persisted content blocks and expose it through agent event APIs without changing the model-visible content
- support reserved `omnara_hidden` and `omnara_display_text` presentation hints in the web transcript
- `omnara_hidden` is used to hide Slack routing context, while `omnara_display_text` is used for preserving canonical mentions and channel references for the model without displaying slack's internal IDs in our web UI
- show message actors in agent conversations and distinguish Slack actors from Omnara users

example:
image


api/openapi/openapi.yaml
```diff
@@ -7068,6 +7068,12 @@ components:
         - ModelOutputStopReasonRefusal
         - ModelOutputStopReasonContentFilter
         - ModelOutputStopReasonError
+    ContentBlockMetadata:
```
cmsparks, line 7071: My openapi metadata refactor hasn't landed yet to share metadata types, but mind doing a type like this so we have metadata size limits:

```
        metadata:
          type: object
          description: Arbitrary key-value metadata. Maximum 16 pairs, keys up to 64 characters, values must be strings of up to 512 characters.
          x-go-type: json.RawMessage
          x-go-type-skip-optional-pointer: true
          maxProperties: 16
          additionalProperties:
            type: string
            maxLength: 512
```

internal/httpapi/public_content_blocks.go
```diff
@@ -164,12 +168,13 @@ func publicModelOutputContentBlocks(
 				return nil, err
 			}
 		case "reasoning":
-			reasoning, err := decodeStoredTextContentBlock(block, index, "reasoning")
+			reasoning, metadata, err := decodeStoredTextContentBlock(block, index, "reasoning")
```
cmsparks, line 171: will assistant/reasoning blocks ever have metadata on them? presumably user supplied input + tool calls would be the only thing that supports metadata?

also, not sure why publicTextContentBlock or publicMediaRefContentBlock don't return metadata above?

cmsparks (approved):
lgtm

internal/httpapi/public_content_blocks.go
asherdale: you're definitely right that assistant/reasoning blocks don't have a metadata producer today, but me and kartik talked about this before i started implementing the PR and thought it would be worth supporting metadata on every persisted content block in this PR for random potential use cases in the future (especially since this is very simple / low cost to implement right now)

[`publicTextContentBlock`](https://github.com/omnara-ai/omnara/blob/8c00892ac455f39cacbcd88a65cbab011c192cbe/internal/httpapi/public_content_blocks.go#L280-L288) and [`publicMediaRefContentBlock`](https://github.com/omnara-ai/omnara/blob/8c00892ac455f39cacbcd88a65cbab011c192cbe/internal/httpapi/public_content_blocks.go#L314-L343) both already return the stored metadata

api/openapi/openapi.yaml
asherdale: good call. i added this shape with string values, `maxProperties: 16`, `maxLength: 512`, and the `json.RawMessage` override. i kept it separate while #305 is still open; once #305 is rebased over this PR, `ContentBlockMetadata` can become a `$ref` to its shared `Metadata` schema

internal/storage/executionstore/content_blocks_store.go
```diff
@@ -100,29 +102,33 @@ func contentBlockArray(contentBlocks json.RawMessage) ([]json.RawMessage, error)
...
 	if kind == "" {
-		return "", nil, errors.New("requires type")
+		return "", nil, nil, errors.New("requires type")
 	}
-	return kind, fields, nil
+	metadata, err := normalizedJSONObject(fields["metadata"], "metadata")
+	if err != nil {
+		return "", nil, nil, err
+	}
+	return kind, metadata, fields, nil
```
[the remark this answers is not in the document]
asherdale: #305 adds OpenAPI 3.1 `propertyNames` validation for this. once it is rebased over this PR, `ContentBlockMetadata` will use its shared `Metadata` schema, so duplicating that validation here is not worthwhile

PR #322, opened 2026-08-07
feat: isolate cluster-managed provider secrets

asherdale:
- Keep cluster-managed credential metadata visible through tenant secret GET and list APIs without exposing plaintext payloads.
- Require organization-owned payload reads to declare and match the expected management kind, preventing cross-boundary decryption.
- Prevent cluster-managed credentials from being reused by tenant model providers, machine pools, machine secret environments, project grants, or tenant-controlled discovery endpoints.
- Preserve default-provider inference by allowing the cluster-managed provider to decrypt only its matching cluster-managed credential.
- Expand regression coverage for visibility, attachment attempts, grant access, environment resolution, and valid default-provider usage.


cmsparks (approved):
lgtm

PR #331, opened 2026-08-10
feat: add safe omnarad uninstall command

asherdale:
- add `omnarad uninstall` with interactive confirmation or `--yes`, complete daemon and agent teardown, managed PATH cleanup, and atomic home retirement
- fail closed for unsafe home targets, unexpected local machine state, and foreign or unverifiable launchd/systemd service definitions
- serialize install, update, supervisor startup, lifecycle commands, and uninstall so concurrent operations cannot resurrect or partially delete daemon state
- document the new command and cover macOS, Linux, process-containment, ownership, recovery, and lifecycle race cases


internal/omnarad/uninstall.go
```diff
@@ -0,0 +1,301 @@
...
+	client := machinedaemon.New(machinedaemon.Config{
+		APIURL:       config.APIURL,
+		MachineToken: config.MachineToken,
+	}, nil, log)
+	if err := client.ReportUninstall(reportCtx, detail); err != nil {
+		log.Warn("report daemon uninstall failed", "error", err)
+	}
+}
+
+func removeManagedPathSymlink(home string) error {
```
cmsparks, line 277: removeManagedPathSymlink is pretty similar to removeSystemdWantsLink (service_systemd.go:368) except resolving the link, maybe we can combine these?

internal/omnarad/service_launchd.go
```diff
@@ -79,9 +79,18 @@ type launchdPlistTemplateData struct {
...
-	registered bool
-	running    bool
-	pid        int
+	definitionPath string
+	registered     bool
+	running        bool
+	pid            int
+}
+
+type managedDaemonStatus struct {
```
cmsparks: do we need the manager attribute on the managedDaemonStatus if we split the types between launchd and system d?

ideally we should instead have a more specific type, maybe `launchdManagedDaemonStatus` and `systemdManagedDaemonStatus`

internal/omnarad/uninstall.go
```diff
@@ -0,0 +1,301 @@
...
+		resultErr = errors.Join(resultErr, daemonLock.Release())
+	}()
+	if config != nil {
+		if err := machinedaemon.StopLocalMachine(
+			ctx,
+			home,
+			config.InstallationID,
+			config.MachineID,
+			"daemon_uninstalled",
+		); err != nil {
```
cmsparks, line 222: nit: we fail to uninstall upon failing to stop. should we instead just soft fail here and force uninstall? because my thought is that we should try our best to uninstall, even if we fail

cmsparks (approved):
just nits, lgtm

internal/omnarad/uninstall.go
asherdale: good call, combined both behind `removeMatchingSymlink` while keeping the existing ownership and exact-target checks

internal/omnarad/service_launchd.go
asherdale: agreed, split these into `launchdManagedDaemonStatus` and `systemdManagedDaemonStatus` and removed the `manager` field

internal/omnarad/uninstall.go
asherdale: yep, changed this to warn and continue uninstalling. when shutdown can't be verified, the command writes a human-readable warning to `stderr`, including process-specific manual cleanup guidance when available

PR #332, opened 2026-08-10
feat: combine agent profiles and agents into one page

ishaansehgal99:
## Summary
- Merge the standalone Agent Profiles tab into the Agents page: the profiles table now stacks above the agents table on `/projects/:projectId/agents`, so the two related concepts live in one place.
- Add a one-click **Launch** button to each profile row (permission-gated, with spinner and error handling) that creates an agent from the profile's current config and navigates to it.
- Hide the search/sort toolbars on both tables by default; they appear only once filtering is active or the list paginates.
- Add an **ID** column (monospace) ahead of **Name** in the agents table, and rename the profiles CTA to **New profile**, inline with the section title.
- Remove the old `/projects/:projectId/agent-profiles` route, `ProjectAgentProfilesPage`, and its sidebar entry (not deployed, so no redirect needed). The profile builder at `/agent-profiles/new` is unchanged and now returns to the combined Agents page.

## Test plan
- [ ] `/projects/:id/agents` shows profiles stacked above agents with correct permission gating (`can_manage` for profile actions, `can_operate` for Launch / New agent)
- [ ] Launch on a profile row creates an agent and navigates to its detail page
- [ ] Toolbars stay hidden for short lists and appear when filtering/paginating
- [ ] Creating a profile via New profile returns to the combined Agents page
- [ ] Old `/agent-profiles` URL 404s; `/agent-profiles/new` still works and highlights the Agents nav item

Made with [Cursor](https://cursor.com)


frontend/apps/web/src/routes/CreateAgentPage.tsx
```diff
@@ -40,95 +34,95 @@ export function CreateAgentPage() {
...
-      const configId =
-        activeTab === 'profile' && selectedProfile
-          ? selectedProfile.current_config_id
-          : (await createAgentConfig.mutateAsync({ source: yaml, source_format: 'yaml' })).id
-      const launch = await createAgent.mutateAsync({
-        profile: activeTab === 'profile' && selectedProfile ? selectedProfile.id : undefined,
-        config: configId,
-        message: draft.message.trim() === '' ? undefined : draft.message,
-      })
-      await navigate({
```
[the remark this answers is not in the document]
ishaansehgal99: Valid catch — addressed. The page now remembers the profile a submit already created (name + YAML + ids) and reuses it when the inputs are unchanged, so retrying after a failed launch no longer creates a duplicate config/profile. If the user edits the name or YAML before retrying, a fresh profile is intentionally created since it's a different config. Fix lands in the next push.

api/openapi/openapi.yaml
```diff
@@ -2448,6 +2448,11 @@ paths:
       description: Items are ordered by created_at descending, then id descending.
       parameters:
         - $ref: '#/components/parameters/ResourceNameFilter'
+        - name: profile
+          in: query
+          description: Return only agents launched from this agent profile.
+          schema:
+            type: string
```
cmsparks: can we make the schema type be an agent profile id. I think we should be able to do this even with a query param
          $ref: '#/components/schemas/AgentProfileID'

migrations/000017_agent_profile_link.sql
```diff
@@ -0,0 +1,40 @@
+-- +goose Up
+
+ALTER TABLE agents
+    ADD COLUMN profile_id uuid;
```
cmsparks: Sort of a preexisting issue, but we inconsistently use profile_id and agent_profile_id in places (I think we use profile_id in the agent_profile_versions table), though I'd lean towards us using agent_profile_id here because it's a bit clearer.

frontend/apps/web/src/components/app-shell/ProjectsNav.tsx
```diff
@@ -48,15 +48,9 @@ export function ProjectsNav() {
...
-                {
-                  to: '/projects/$projectId/agent-profiles' as const,
-                  path: `${projectRoot}/agent-profiles`,
-                  label: 'Agent Profiles',
-                  prefix: true,
-                },
                 {
                   to: '/projects/$projectId/agents' as const,
-                  path: `${projectRoot}/agents`,
+                  path: `${projectRoot}/agent`,
```
cmsparks: nit: i feel like the path should be plural because the other ones are plural

cmsparks (approved):
lgtm, just minor nits

PR #334, opened 2026-08-11
chore: Remove most instances of sleep in tests. Add linter to restrict sleep calls

cmsparks:


internal/machinedaemon/runner_protocol.go
```diff
@@ -39,18 +39,28 @@ const (
...
+	runnerStageActionCommitted   runnerStage = "action_committed"
+	runnerStageFenceWriterParked runnerStage = "fence_writer_parked"
+)
+
 type runnerRequest struct {
 	SupervisorToken      string          `json:"supervisor_token,omitempty"`
 	SupervisorInstanceID string          `json:"supervisor_instance_id,omitempty"`
 	Method               runnerMethod    `json:"method"`
 	Payload              json.RawMessage `json:"payload,omitempty"`
+	NotifyStages         bool            `json:"notify_stages,omitempty"`
```
asherdale: could we avoid putting these test stages on the real runner protocol? `action_committed` is already observable by waiting on the SQLite row. for `fence_writer_parked`, `TestMain` already owns the child startup, so it could install a small hook that writes to a named pipe passed in the child env. that would let us drop `NotifyStages`, the multi-frame reader, and the stage plumbing across `runner_protocol.go`, `runner_ipc_client.go`, and `runner_server.go`

internal/machinedaemon/report_outbox.go
```diff
@@ -107,6 +107,9 @@ func (c *Client) replayReportOutbox(
 				break
 			}
 			delete(forced, report.ID)
+			if c.reportSettled != nil {
```
asherdale: could we avoid the test-only `reportSettled` hook by extracting one pass of `replayReportOutbox` into a synchronous helper? the test could call it twice and then assert the persisted report states directly. that’d keep this sleep-free without adding a callback to `Client` that runs in the prod outbox loop

asherdale (approved):

PR #337, opened 2026-08-11
feat: cache SigV4 MCP role credentials

asherdale:
- Keeps one worker-scoped cache of AWS SDK credential providers for SigV4 MCP servers that assume roles, avoiding a fresh STS session for every connection.
- Keys cached providers by secret, version, and region; secret rotation replaces stale entries and a 1,024-entry cap bounds process memory.
- Leaves static AWS credentials uncached and performs project-scoped secret authorization before every cache lookup.
- Continues retrieving credentials and generating a fresh SigV4 signature for every MCP request while the SDK handles expiry-aware refresh and single-flight retrieval.

### DIFF
--- a/internal/mcp/sigv4.go
+++ b/internal/mcp/sigv4.go
@@ -15,6 +15,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/credentials"
 	"github.com/aws/aws-sdk-go-v2/credentials/stscreds"
 	"github.com/aws/aws-sdk-go-v2/service/sts"
+	"github.com/hashicorp/golang-lru/v2/simplelru"
 	"github.com/omnara-ai/omnara/internal/secrets"
 	"github.com/omnara-ai/omnara/internal/storage"
 )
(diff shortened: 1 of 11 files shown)


internal/mcp/sigv4.go
```diff
@@ -15,13 +16,77 @@ import (
 	"github.com/aws/aws-sdk-go-v2/credentials/stscreds"
 	"github.com/aws/aws-sdk-go-v2/service/sts"
 	"github.com/omnara-ai/omnara/internal/secrets"
+	"github.com/omnara-ai/omnara/internal/storage"
 )
 
+const maxSigV4CredentialCacheEntries = 1_024
+
+type SigV4CredentialCache struct {
```
cmsparks, line 25: optional, but we could use: https://github.com/hashicorp/golang-lru

this is a pretty standard generic lru package that a lot of ppl use in golang

cmsparks (approved):

internal/mcp/sigv4.go
asherdale: good call, added this in the latest commit

PR #339, opened 2026-08-11
feat: harden database evolution and bearer credentials

ksarangmath:
## What this changes

- Prevents database migrations from being changed after they have shipped in a release. Migrations that have not shipped yet can still be edited.
- Makes database upgrades safer with a PostgreSQL 18 check, migration timeouts, and sensible connection-pool defaults that still respect explicit configuration.
- Gives personal access tokens, organization API keys, and machine credentials one consistent, secure format.
- Keeps token details private to the server, so clients only need to store and send the value they receive.
- Centralizes how authenticated users, organization keys, and machines are represented and removes unused authentication concepts.
- Adds coverage for database upgrades, API compatibility, credentials, and deployment behavior.

## Important rollout notes

This is a clean token-format change. Existing personal access tokens, organization API keys, and machine credentials will stop authenticating and must be replaced. Their database records will remain until revoked and will continue counting toward token limits.

Avoid running old and new API versions at the same time during this deployment. Because the API, worker, and maintenance services can all issue machine credentials, deploy those services together for this cutover.

Run the migration image from this release before starting its API, worker, and maintenance services. Service readiness checks confirm that PostgreSQL and Redis are reachable; the migration job remains responsible for upgrading the database.


Makefile
```diff
@@ -186,6 +220,44 @@ migration-check:
 			expected=$$((expected + 1)); \
 		done; \
 	done
+	$(MIGRATION_CHECK) check
+
+migration-compat-check:
+	$(MIGRATION_CHECK) compare-releases $(if $(MIGRATION_RELEASE_REF_ROOT),--release-ref-root "$(MIGRATION_RELEASE_REF_ROOT)")
+
+compatibility-check: ## Check API and released migrations after syncing with origin/main
```
asherdale, line 208: nonblocking nit: `compatibility-check` (and `openapi-compat-fixture-check` too) contains enough shell control flow that they may be easier to test and maintain under `tools/ci`, which would set a good standard for the long-term hygiene and maintenance of the makefile. nonblocking though and completely up to you

api/openapi/openapi.yaml
```diff
@@ -5938,8 +5938,8 @@ components:
     bearerAuth:
       type: http
       scheme: bearer
-      bearerFormat: Omnara personal access token
-      description: A user's personal access token prefixed omnara_pat_.
+      bearerFormat: Omnara personal or organization access token
+      description: An opaque Omnara personal or organization bearer token. Clients must not parse or validate its internal format.
```
asherdale: should we remove `Clients must not parse or validate its internal format.` from all five occurrences? calling the credential `opaque` already communicates this, so the repeated sentence just adds noise

internal/bearertoken/token.go
```diff
@@ -0,0 +1,173 @@
...
+)
+
+const (
+	alphabet       = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
+	secretLength   = 43
+	checksumLength = 6
+
+	personalAccessPrefix = "omnara_pat_v1_"
+	organizationPrefix   = "omnara_org_v1_"
+	daemonPrefix         = "omnara_daemon_v1_"
```
asherdale: should we build these prefixes from the corresponding `Kind*` constants? right now `pat`, `org`, and `daemon` are used as literals twice which feels unnecessary (and hypothetically could drift)

internal/bearertoken/token.go
```diff
@@ -0,0 +1,173 @@
...
+		return "", fmt.Errorf("%w: unsupported kind %q", ErrInvalid, kind)
+	}
+	if err := validateBase62(secret, secretLength, "secret"); err != nil {
+		return "", err
+	}
+	body := prefix + secret
+	return body + string(checksumSeparator) + encodeChecksum(crc32.ChecksumIEEE([]byte(body))), nil
+}
+
+// Parse validates a token and returns its credential kind.
```
asherdale: should we remove the comments that only restate simple/readable code: the package comment above, this `Parse` comment, and the one on `WithDefaultApplicationName`? they add noise without explaining anything non-obvious

asherdale (approved):
lgtm, just pointed out a few nits

Makefile
ksarangmath: Addressed in 09170fec1. The Makefile now keeps only the entry points; the compatibility orchestration and OpenAPI fixture control flow live in focused scripts under tools/ci with the same behavior.

api/openapi/openapi.yaml
ksarangmath: Addressed in 205c7f790. The five OpenAPI descriptions now retain “opaque” without repeating the parsing instruction; the authentication guide remains the single plain-language explanation.

internal/bearertoken/token.go
ksarangmath: Addressed in 85c7d8455. Each fixed prefix is now derived at compile time from its corresponding Kind constant, while the token format and golden vectors remain unchanged.

internal/bearertoken/token.go
ksarangmath: Addressed in 649cd2ccf. I removed the three narration-only comments and audited every other handwritten comment added by this PR; the remaining comments explain non-obvious security, compatibility, locking, release-boundary, or configuration invariants.

PR #341, opened 2026-08-12
feat: add organization resource limit overrides

asherdale:
- add a validated resource_limit_overrides JSONB field to organizations with optional overrides for all 14 existing org-resolved resource limits
- pair each override key with its application default and fall back to that default when an organization has no override
- resolve overrides inside the existing transactional admission paths while preserving idempotency, concurrency safety, scope boundaries, and fixed user-owned limits
- raise the reviewed default capacities to 1,000 or 10,000 where applicable while retaining the existing daemon-token and process limits

- supersede the original JSONB design with nullable, typed columns in a dedicated org_resource_limit_overrides table
- define defaults in the default_resource_limits SQL view and combine them with overrides through effective_resource_limits, so a missing row or NULL column uses the default
- map limit resolution for a deleted organization to NotFound and keep the BYO machine project-ID request cap independent from the project resource limit

### DIFF
--- a/internal/dbmigrate/postgres_integration_test.go
+++ b/internal/dbmigrate/postgres_integration_test.go
@@ -66,7 +66,7 @@ func TestPostgresStoredOrgScopeColumnsMatchOwnershipBoundaries(t *testing.T) {
 	defer cancel()
 
 	_, db := openPostgresMigrationTestDB(t, ctx)
-	const expected = "agent_configs,agent_machine_bindings,agents,configured_model_revisions,configured_models,daemon_runtimes,integration_installs,machine_daemon_tokens,machine_online_intervals,machine_pools,machines,model_call_contexts,model_provider_configs,org_api_keys,org_invitations,org_managed_work_admission,org_memberships,process_actions,processes,project_machine_grants,project_machine_pool_grants,project_memberships,project_model_grants,projects,secret_grants,secret_oauth_refresh_leases,secret_versions,secrets,skill_grants,skills"
+	const expected = "agent_configs,agent_machine_bindings,agents,configured_model_revisions,configured_models,daemon_runtimes,integration_installs,machine_daemon_tokens,machine_online_intervals,machine_pools,machines,model_call_contexts,model_provider_configs,org_api_keys,org_invitations,org_managed_work_admission,org_memberships,org_resource_limit_overrides,process_actions,processes,project_machine_grants,project_machine_pool_grants,project_memberships,project_model_grants,projects,secret_grants,secret_oauth_refresh_leases,secret_versions,secrets,skill_grants,skills"
 	var actual string
 	if err := db.QueryRowContext(ctx, `
 SELECT coalesce(string_agg(column_info.table_name, ',' ORDER BY column_info.table_name), '')
--- a/internal/harness/tools/process_tools.go
+++ b/internal/harness/tools/process_tools.go
@@ -134,10 +134,7 @@ func startCommand(
 				return failProcessTransaction(
 					err,
 					processToolErrorAgentLimitReached,
-					fmt.Sprintf(
-						"this agent already has %d unfinished processes; use list_processes and stop_process before starting another",
-						executionstore.MaxNonTerminalProcessesPerAgent,
-					),
+					"this agent has reached its unfinished process limit; use list_processes and stop_process before starting another",
 					false,
 					"",
 					"",
--- a/internal/httpapi/default_model_provider_org_integration_test.go
+++ b/internal/httpapi/default_model_provider_org_integration_test.go
@@ -152,7 +152,7 @@ func TestConcurrentOrganizationCreationCommitsOnlyOneFinalOwnedSlot(t *testing.T
 	)
 	store := integrationStoreForHandler(t, handler)
 	user, token := createOrgRouteUser(t, pool, store, "default-provider-final-slot")
-	for index := 0; index < 99; index++ {
+	for index := 0; index < identitystore.MaxOwnedOrgsPerUser-1; index++ {
 		if _, err := store.Organizations().CreateOrgForUser(ctx, orglifecycle.CreateOrgForUserInput{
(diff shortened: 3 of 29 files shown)


migrations/000018_org_resource_limit_overrides.sql
```diff
@@ -0,0 +1,32 @@
+-- +goose Up
+
+ALTER TABLE orgs
+ADD COLUMN resource_limit_overrides jsonb NOT NULL DEFAULT '{}'::jsonb,
```
cmsparks: I don't love this datamodel IMO, if we're going to store it in the DB, IMO I'd do two things:
* strictly type it in columns in a separate table
* store defaults in the DB (either in a table or in a view)

```sql
-- claude made up numbers here. the defaults also could alternatively be a separate table which we update
CREATE VIEW default_resource_limits AS
SELECT
    5   AS max_active_projects_per_org,
    25  AS max_pending_org_invitations_per_org,
    10  AS max_active_org_api_keys_per_org,
    5   AS max_active_tenant_model_provider_configs_per_org,
    10  AS max_active_configured_models_per_provider,
    20  AS max_agent_configs_per_project,
    20  AS max_active_agent_profiles_per_project,
    10  AS max_active_agents_per_project,
    50  AS max_active_tenant_secrets_per_owner,
    50  AS max_active_skills_per_owner,
    3   AS max_active_tenant_machine_pools_per_org,
    10  AS max_live_machines_per_org,
    5   AS max_active_byo_daemon_tokens_per_machine,
    10  AS max_non_terminal_processes_per_agent;

CREATE VIEW effective_resource_limits AS
SELECT
    o.id AS org_id,
    COALESCE(l.max_active_projects_per_org,                      d.max_active_projects_per_org)                      AS max_active_projects_per_org,
    COALESCE(l.max_pending_org_invitations_per_org,              d.max_pending_org_invitations_per_org)              AS max_pending_org_invitations_per_org,
    COALESCE(l.max_active_org_api_keys_per_org,                  d.max_active_org_api_keys_per_org)                  AS max_active_org_api_keys_per_org,
    COALESCE(l.max_active_tenant_model_provider_configs_per_org, d.max_active_tenant_model_provider_configs_per_org) AS max_active_tenant_model_provider_configs_per_org,
    COALESCE(l.max_active_configured_models_per_provider,        d.max_active_configured_models_per_provider)        AS max_active_configured_models_per_provider,
    COALESCE(l.max_agent_configs_per_project,                    d.max_agent_configs_per_project)                    AS max_agent_configs_per_project,
    COALESCE(l.max_active_agent_profiles_per_project,            d.max_active_agent_profiles_per_project)            AS max_active_agent_profiles_per_project,
    COALESCE(l.max_active_agents_per_project,                    d.max_active_agents_per_project)                    AS max_active_agents_per_project,
    COALESCE(l.max_active_tenant_secrets_per_owner,              d.max_active_tenant_secrets_per_owner)              AS max_active_tenant_secrets_per_owner,
    COALESCE(l.max_active_skills_per_owner,                      d.max_active_skills_per_owner)                      AS max_active_skills_per_owner,
    COALESCE(l.max_active_tenant_machine_pools_per_org,          d.max_active_tenant_machine_pools_per_org)          AS max_active_tenant_machine_pools_per_org,
    COALESCE(l.max_live_machines_per_org,                        d.max_live_machines_per_org)                        AS max_live_machines_per_org,
    COALESCE(l.max_active_byo_daemon_tokens_per_machine,         d.max_active_byo_daemon_tokens_per_machine)         AS max_active_byo_daemon_tokens_per_machine,
    COALESCE(l.max_non_terminal_processes_per_agent,             d.max_non_terminal_processes_per_agent)             AS max_non_terminal_processes_per_agent
FROM orgs o
CROSS JOIN default_resource_limits d
LEFT JOIN org_resource_limit_overrides l ON l.org_id = o.id;
```

cmsparks: This has the added benefit of allowing us to move the enforcement of our resource limits entirely to the queries (and we get to rip out a ton of golang code).

For example:

```sql
-- name: CreateProject :one
INSERT INTO projects (org_id, name)
SELECT sqlc.arg(org_id), sqlc.arg(name)
WHERE (
    SELECT count(*) FROM projects
    WHERE org_id = sqlc.arg(org_id) AND deleted_at IS NULL
) < (
    SELECT max_active_projects_per_org
    FROM effective_resource_limits
    WHERE org_id = sqlc.arg(org_id)
)
RETURNING *;
```

Technically this version contains a race condition which may incidentally allow extra rows, but 1) probably fine in this scenario 2) there are other solutions that don't result in races

cmsparks (commented):
can discuss tmrw, but i'd lean towards a different schema here (which probably results in substantial changes to the PR)

cmsparks (approved):

cmsparks:
lgtm, not gonna bikeshed

migrations/000018_org_resource_limit_overrides.sql
cmsparks: from discussion: 
```sql
CREATE TABLE tier_defaults
    tier_id
    limit_a NOT NULL
    limit_b NOT NULL
    limit_c NOT NULL

CREATE TABLE org_resource_limit_overrides 
  org_id FK -> orgs
  tier_id FK -> tier_defaults
  limit_a NULLABLE
  limit_b NULLABLE
  limit_c NULLABLE

SELECT 
 org_id
 tier_id
 coalesce(o.limit_b, d.limit_b)
from org_resource_limit_overrides
where org_id = blah
join tier_id
```

asherdale: good call, will move to a new data model with a table for overrides and a view for the default limits

asherdale: will keep enforcement in Go to avoid race conditions and retain the desired idempotency behavior

cmsparks (approved):
lgtm!

PR #343, opened 2026-08-12
feat: Onboarding flow with agent templates and first-agent CTA

ishaansehgal99:
## Summary

Makes onboarding for a fresh organization smooth: new users land on a clear path from "empty org" to a running agent, with docs one click away.

### Agent templates on the New Agent page
- Adds three prebuilt templates — **General agent**, **Deep researcher**, and **Structured extractor** — available from a compact "Templates" dropdown in the page header (`AgentTemplateMenu`). The blank form remains the default starting state.
- All templates share the same built-in toolset (process/machine management, `ask_question`, `web_search`, `web_fetch`) and auto-select the org's default machine pool.
- Supports `?template=` deep links: the route waits for the tool catalog and machine pools to load, then prefills the form via state initializers (no `useEffect` state syncing); changing the search param remounts the form with the new template.
- Refactors the oversized `CreateAgentPage` into a thin route wrapper plus a `CreateAgentForm` component; shared config helpers moved to `agentConfigBasicSerialization`.

### First-agent CTA on Overview
- Fresh orgs (no agents, no agent profiles, at least one manageable project) see a `FirstAgentCard` banner: a short pitch, a "New agent" button deep-linking to the general template, and a footnote link to the API docs for users who prefer code.

### Recent agents and profiles on Overview
- Established orgs get a "Recent agents" section as the Overview content: the 5 most recently active agents across projects (name, project, model, last active), each row linking to the agent. A new internal `getOrgOverview` endpoint returns the caller's visible projects plus the most recent agents and agent profiles across every project they can read, so the page loads in a single request with per-project authorization applied server-side.
- When no agents are running but profiles exist, the section swaps to "Recent profiles" with an explainer line — launching stays a deliberate action taken from the profile page, covering the profile-created-but-never-launched state that previously had no CTA on Overview.
- A "New agent" header button links straight to the only manageable project, or offers a project picker when there are several; hidden for viewers.
- Removes the "Recent pages" table and its localStorage tracking (`recent-pages.ts` and the `AppShell` recorder), since the agents/profiles view replaces it as the "resume work" surface.

### Docs discoverability
- Adds a persistent "Documentation" link (docs.omnara.com) in the sidebar footer, styled like the rest of the nav with an external-link indicator.

The docs-site theme alignment was split into #344.

## Test plan
- [ ] Fresh org: Overview shows the first-agent card
- [ ] Org with profiles but no agents: Overview shows "Recent profiles" rows linking to profile pages
- [ ] Org with agents: Overview shows "Recent agents" across projects sorted by last activity; rows navigate to the agent
- [ ] "New agent" header button: direct link with one manageable project, project picker with several, hidden for viewers
- [ ] "New agent" button on the first-agent card prefills the General agent template (name, instructions, tools, default machine pool)
- [ ] Templates dropdown applies each template; blank form remains the default without a template
- [ ] `?template=` deep links work after catalog/pool load; invalid ids fall back to blank
- [ ] Sidebar "Documentation" link opens docs.omnara.com in a new tab


frontend/apps/web/src/routes/Overview.tsx
```diff
@@ -1,114 +1,45 @@
...
+  const agentsQuery = useAgents(activeOrg.id, onboardingProject?.id ?? '', {
+    pageSize: 1,
+    enabled: onboardingProject != null,
+  })
+  const overviewPending =
+    onboardingProject != null && (profilesQuery.isPending || agentsQuery.isPending)
+  const showOnboarding =
+    onboardingProject != null &&
+    profilesQuery.data?.pages[0]?.data.length === 0 &&
+    agentsQuery.data?.pages[0]?.data.length === 0
```
cmsparks: nit: if you have multiple projects, but no agents, this card doesn't show

frontend/apps/web/src/components/agents/agentTemplates.ts
```diff
@@ -0,0 +1,110 @@
...
+  instruction: `You extract structured data from unstructured text. Given raw input (emails, PDFs, logs, transcripts, scraped HTML) and a target JSON schema:
+
+1. Read the schema first. Note required vs optional fields, enums, and format constraints (dates, currencies, IDs). The schema is the contract — never emit a key it doesn't define.
+2. Scan the input for each field. Prefer explicit values over inferred ones. If a required field is genuinely absent, use null rather than guessing.
+3. Normalize as you extract: trim whitespace, coerce dates to ISO 8601, strip currency symbols into numeric + code, collapse enum synonyms to their canonical value.
+4. Emit a single JSON object (or array, if the schema is a list) that validates against the schema. No prose, no markdown fences — just the JSON.
+
+When the input is ambiguous, pick the most conservative interpretation and note the ambiguity in a top-level "_extraction_notes" field only if the schema allows additionalProperties.`,
+  firstMessagePlaceholder: 'Paste raw text plus the JSON schema you want extracted',
+}
```
cmsparks, line 68: I think there's two things I think we could improve about the templates:
1) They should be more useful because the hard part of an agent isn't the system prompt IMO
2) They should be more prominent, because currently they're kinda hidden in the corner under a dropdown

For point 1, we should add some model + machine configuration, skills, mcp, etc. I don't think we have to do all of that right now, but I think at a minimum we should preselect the model and pre-select a machine pool when selecting a template. We can add richer config later (skills + mcp)

For 2, I think we should have a card for each template in the FirstAgentCard where if you click it autofills and sends you to the builder screen, as opposed to the user needing to find the template dropdown.

frontend/apps/web/src/components/overview/RecentAgents.tsx
```diff
@@ -0,0 +1,216 @@
...
+}
+
+interface ProfileRow {
+  profile: AgentProfile
+  project: VisibleProject
+}
+
+export function RecentAgentsSection({
+  orgId,
+  projects,
```
cmsparks: projects probably shouldn't be drilled in as a prop here

frontend/apps/web/src/components/overview/RecentAgents.tsx
```diff
@@ -0,0 +1,216 @@
...
+    .sort((a, b) => b.agent.updated_at.localeCompare(a.agent.updated_at))
+    .slice(0, rowLimit)
+
+  const isPending = results.some((result) => result.isPending)
+  const isError = rows.length === 0 && results.some((result) => result.isError)
+
+  const agentsEmpty = !isPending && !isError && rows.length === 0
+  const profileResults = useQueries({
+    queries: readableProjects.slice(0, maxProjects).map((project) => ({
+      ...listAgentProfilesOptions({
```
cmsparks: We're doing 2 n+1 10 queries every time here up to 20 queries. And this is after depending on the prop drilled readableProjects. 

I think it'd be better if we just created a dedicated `${org_id}/overview/` route which does these queries for us. 

Right now it'd just have recent agents + agent profiles, but we can just change it to add whatever data we want from the backend later.

To prevent end users from using or seeing the endpoint in the docs we could mark it as `x-hidden: true`

frontend/apps/web/src/routes/Overview.tsx
```diff
@@ -1,114 +1,45 @@
...
-      ]
-    })
-    .slice(0, 5)
+  const firstProject = projects.length === 1 ? projects[0] : undefined
+  const onboardingProject = firstProject?.access.can_manage ? firstProject : undefined
+  const profilesQuery = useAgentProfiles(activeOrg.id, onboardingProject?.id ?? '', {
+    pageSize: 1,
+    enabled: onboardingProject != null,
+  })
+  const agentsQuery = useAgents(activeOrg.id, onboardingProject?.id ?? '', {
```
cmsparks: from my commment in RecentAgents, we can probably omit these two queries and a lot of conditional logic if we have a dedicated overview route

frontend/apps/web/src/components/agents/agentTemplates.ts
cmsparks: Oh I think the intent is that it sets the machine pool, but if my read is right, I think we we don't actually grant the pool or the model, so users won't have access to the default pool/model in the first place.

IMO what we should do is:
1) automatically grant the cluster managed pool and some cluster managed model (we should preset this to a good model) to the project if it has not been granted yet
2) use the previously granted pool

frontend/apps/web/src/routes/CreateAgentPage.tsx
```diff
@@ -83,181 +39,18 @@ export function CreateAgentPage() {
...
+    <CreateAgentForm
+      orgId={activeOrg.id}
+      orgName={activeOrg.name}
+      projectId={projectId}
+      projectName={project.name}
+      catalog={catalog}
+      defaultPool={defaultPool}
+      templatesReady={templatesReady}
+      initialTemplate={linkedTemplate}
+    />
```
cmsparks: If we're refactoring this component to just be a thin wrapper, we probably don't need to be drilling all this state in. 

orgId, orgName, projectId, and projectName can be derived by useActiveOrg and useProjectPage inside of the create agent form component (this is somewhat of a preexisting issue)

However these attributes, 
```
      catalog={catalog}
      defaultPool={defaultPool}
      templatesReady={templatesReady}
      initialTemplate={linkedTemplate}
```

Are all data which is only used inside of the CreateAgentForm component, so we should use the hooks inside of that component.

cmsparks (commented):
lgtm, some changes that could improve ux and some react structure changes

cmsparks (approved):

frontend/apps/web/src/components/overview/RecentAgents.tsx
ishaansehgal99: this is good feedback. done, added GET /orgs/{orgID}/overview marked x-hidden: true. so whole page is now one request as opposed to like 20

ishaansehgal99: also endpoint easy to extend with more overview data later if we want to edit it

frontend/apps/web/src/components/overview/RecentAgents.tsx
ishaansehgal99: fixed as part of the overview endpoint

frontend/apps/web/src/routes/Overview.tsx
ishaansehgal99: yep, both probe queries and the enable/pending gymnastics are gone

frontend/apps/web/src/routes/Overview.tsx
ishaansehgal99: fixed

frontend/apps/web/src/routes/CreateAgentPage.tsx
ishaansehgal99: good call - moved all of this out of the route

frontend/apps/web/src/components/agents/agentTemplates.ts
ishaansehgal99: so there is no autogranting currently you are right. we should address this in future pr. that said the default machine pool is autogranted to the default project and upon clicking the overview page's new agent button automatically gets prefilled

frontend/packages/react/src/domains/agents.ts
```diff
@@ -56,9 +57,14 @@ export function useCreateAgent(orgID: string, projectID: string) {
...
-          queryKey: listAgentsQueryKey({ path: { orgID, projectID }, client }),
-        })
+        await Promise.all([
+          queryClient.invalidateQueries({
+            queryKey: listAgentsQueryKey({ path: { orgID, projectID }, client }),
+          }),
+          queryClient.invalidateQueries({
+            queryKey: getOrgOverviewQueryKey({ path: { orgID }, client }),
+          }),
+        ])
```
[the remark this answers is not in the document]
ishaansehgal99: Fixed in 8200b63 — `useCreateProject` now also invalidates the org overview query.

frontend/apps/web/src/components/agents/CreateAgentForm.tsx
```diff
@@ -0,0 +1,44 @@
...
+  if (linkedTemplate && !templatesReady) return <FullPageSpinner />
+
+  return (
+    <CreateAgentFormView
+      key={linkedTemplate?.id ?? ''}
+      catalog={catalog}
+      defaultPool={defaultPool}
+      defaultModel={defaultModel}
+      templatesReady={templatesReady}
+      initialTemplate={linkedTemplate}
```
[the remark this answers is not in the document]
ishaansehgal99: Intentional. Pool/model grants are optional prefill and 403 for users without access_manage. We wait for the queries to settle, then mount with whatever loaded instead of blocking the page. Catalog/grant failures leave those fields empty for the builder.

PR #345, opened 2026-08-12
feat: unify app and docs favicon

asherdale:
- add the circular Omnara favicon used by the marketing site to the app frontend
- update the docs favicon to the same asset so `app.omnara.com`, `docs.omnara.com`, and `omnara.com` share consistent browser-tab branding

image


ishaansehgal99 (approved):
From Carolina hills to Tenderloin streets,
a monk came west where ambition meets.
He left the temple, kept the climb—
all back, no legs, enlightenment takes time.

PR #346, opened 2026-08-12
feat: add help link to user menu

asherdale:
- adds a `Need help?` action directly above log out so support is discoverable from every authenticated page
- opens the existing [Omnara support page](https://docs.omnara.com/support) in a new tab without interrupting console work

Normal state:
image

State on cursor hover:
image


ksarangmath (approved):
LGTM

PR #348, opened 2026-08-12
feat: Improve agent edit UI

cmsparks:
API changes:
* add name attribute to update profile
* attach machine binding and mcp state to GET /agent/agent_id

Frontend changes:
* agent chat:
  * info sidebar
  * config editor
* agent profile
  * config editor (builder)
  * integrations tab
  * misc cleanup


frontend/apps/web/src/components/agents/agentConfigBasicDeserialization.ts
```diff
@@ -0,0 +1,246 @@
...
+} from '@/components/machines/machineOverrides'
+
+/**
+ * Best-effort inverse of serializeBasicConfig. Returns a builder draft only
+ * when re-serializing it yields an equivalent parsed document, so the builder
+ * can never silently change or drop configuration data it doesn't understand
+ * (unknown fields, provider options overlays, non-built-in tools). Formatting
+ * and comments are not preserved: saving from the builder normalizes them.
+ */
+export function deserializeBasicConfig(source: string): BasicConfig | null {
```
cmsparks: this file is _kinda_ slop, but I think it's sufficient for doing some best effort parsing. The goal is that I should be able to swap between yaml and the builder and the config doesn't get completely broken

cmsparks: swapping this to document based yaml parsing so it's not slop

api/openapi/openapi.yaml
```diff
@@ -7569,10 +7587,15 @@ components:
...
     UpdateAgentProfileRequest:
+      description: >-
+        Updates exactly one of the profile name or the current config. A config
+        update must supply both config and expected_current_config_id. Renames
+        do not accept an Idempotency-Key: they have no replay ledger.
       type: object
       additionalProperties: false
-      required: [config, expected_current_config_id]
       properties:
+        name:
```
asherdale: should we expose profile renames on the profile resource itself, e.g. `PATCH /agent-profiles/{agentProfileID}`, instead of sending `{name: ...}` to `/agent-profiles/{agentProfileID}/config`? the current route makes a metadata update look like a config update and gives one endpoint two unrelated payload and idempotency contracts, including accepting `Idempotency-Key` for config changes but explicitly rejecting it for renames (which seems smelly)

another example is that a request that only carries a name update still loads the current config of the profile and includes it in the response, which seems unnecessary/unrelated (from the frontend's perspective it definitely makes more sense, but for an API caller the coupling seems more unnecessary, at least in my eyes)

frontend/apps/web/src/components/agents/useAgentConfigEditor.ts
```diff
@@ -0,0 +1,82 @@
...
+  yamlDiverged,
+} from '@/components/agents/agentConfigModeMachine'
+import {
+  createBasicConfigSession,
+  useAgentBuilderForm,
+} from '@/components/agents/useAgentBuilderForm'
+
+export type AgentConfigEditorState = ReturnType<typeof useAgentConfigEditor>
+
+export function useAgentConfigEditor({
```
cmsparks, line 16: there's a bit of slop in these editor hooks (some unnecessary state), but it should be pretty failure-proof due to the tests

frontend/apps/web/src/components/agents/useAgentBuilderForm.ts
```diff
@@ -0,0 +1,398 @@
...
+      wire.initial_num_machines = Number(source.initialNumMachines)
+    }
+    if (source.maxMachines !== '') wire.max_machines = Number(source.maxMachines)
+    if (source.machineCpu !== '') wire.machine_cpu = Number(source.machineCpu)
+    if (source.machineMemoryMb !== '') wire.machine_memory_mb = Number(source.machineMemoryMb)
+    const optionsOverlay = isMachinePoolProvider(source.provider)
+      ? providerOptionsOverlay(
+          source.provider,
+          source.providerOptions,
+          source.managementKind === 'cluster',
```
asherdale: should we preserve existing resource/location overlay keys when serializing a cluster-managed source? the builder hides those fields, but the backend allows them and the draft retains them. changing an unrelated field like `cwd` rebuilds the row with `clusterManaged` true, which drops the existing `image` or location override. basically the hidden values get deleted by unrelated edits

frontend/apps/web/src/routes/AgentView.tsx
```diff
@@ -41,52 +61,110 @@ export function AgentView() {
...
+            canOperate={canOperate}
+          />
+        </div>
       </div>
-    </div>
+      <AgentSidebar
+        orgId={activeOrg.id}
+        projectId={projectId}
+        agent={agent}
+        machineIds={data.machine_ids}
```
asherdale, line 163: should we refresh these sidebar values while they are transitional? a pool machine is attached before it becomes `active`, so the row's `GET /machines/{id}` initially returns 404. 4xx queries are not retried, and nothing invalidates that machine query when provisioning completes

similarly, `mcp_connections` comes from a one-shot agent query, so `initializing` can remain "Connecting" after it becomes `ready` or `failed`. we could poll or invalidate until these states settle

frontend/packages/react/src/domains/agent-profiles.ts
asherdale: should we write `_data` to the profile detail cache here before invalidating? the editor only clears after `profile.current_config_id` changes. if the refetch fails, the save has committed but the UI stays dirty with the old id, and retrying 409s

internal/storage/queries/agent_runtime.sql
```diff
@@ -1,18 +1,36 @@
 -- name: InsertAgent :one
-INSERT INTO agents(
-    org_id, project_id, state, name, agent_profile_id, current_config_id,
-    idempotency_key, created_at, updated_at
+-- @sqlc-vet-disable configured-models-deleted-at model-provider-configs-deleted-at
```
asherdale, line 2: minor nit: for consistency's sake, should we add a rationale next to these three vet suppressions? every existing `@sqlc-vet-disable` on `main` has a comment explaining the reason for it, which feels like a worthwhile standard to keep up

internal/httpapi/agent_routes.go
```diff
@@ -913,12 +957,52 @@ func (s *Server) startPoolMachineDeletion(parent context.Context, machines []exe
...
+func (s *Server) currentAgentResponse(
+	ctx context.Context,
+	record executionstore.AgentRecord,
+) (openapi.GetAgentResponse, error) {
 	agent, err := publicAgentResponseFromRecord(record)
 	if err != nil {
 		return openapi.GetAgentResponse{}, err
 	}
-	return openapi.GetAgentResponse{Agent: agent}, nil
+	records, err := s.store.Execution().ListAgentMachineBindings(ctx, record.ProjectID, record.ID)
```
asherdale, line 982: should archive and idempotent create avoid these follow-up reads? both operations have already succeeded by this point, so a failed read returns an error even though the requested mutation completed

frontend/apps/web/src/routes/CreateAgentPage.tsx
```diff
@@ -63,11 +63,27 @@ export function CreateAgentPage() {
...
     message: '',
-    basicConfig: createEmptyBasicConfigDraft(),
     status: idle,
   })
   const [pendingAction, setPendingAction] = useState<SubmitAction | null>(null)
   const savedProfile = useRef<SavedProfile | null>(null)
+  const [session, setSession] = useState(() => createBasicConfigSession(''))
+  const form = useAgentBuilderForm(session)
+  const [builderYaml, setBuilderYaml] = useState('')
+  if (!form.blocked && builderYaml !== form.yaml) setBuilderYaml(form.yaml)
```
asherdale: should `builderYaml` keep tracking `form.yaml` while the form is incomplete? this freezes it at the last valid YAML, so switching to YAML and submitting from there can silently omit the latest builder edits

frontend/apps/web/src/components/agents/agentConfigBasicExtract.ts
```diff
@@ -0,0 +1,177 @@
...
+  machine_name: z.string(),
+  cwd: z.string().optional(),
+  env_overlay: overlay,
+  secret_env_overlay: overlay,
+})
+
+const poolEntry = z.strictObject({
+  machine_pool_name: z.string(),
+  initial_num_machines: count,
+  max_machines: count,
```
asherdale: should we use nonnegative validation for these two fields and add round-trip coverage for `0`?

both `initial_num_machines` and `max_machines` validly allow `0` in an agent config, but this shared positive-only schema rejects both, so otherwise valid configs can’t be shown in the builder

frontend/apps/web/src/components/agents/agentConfigBasicExtract.ts
```diff
@@ -0,0 +1,177 @@
...
+  machine_cpu: count,
+  machine_memory_mb: count,
+  machine_provider_options_overlay: overlay,
+  cwd: z.string().optional(),
+  env_overlay: overlay,
+  secret_env_overlay: overlay,
```
asherdale, line 44: similar to the previous comment, `env_overlay` and `secret_env_overlay` allow `null` values to remove inherited pool values, and the shared row types already model that state. but this string-only schema rejects those valid configs

frontend/apps/web/src/components/agents/agentConfigBasicExtract.ts
```diff
@@ -0,0 +1,177 @@
...
+  url: z.string(),
+  permission,
+  default_enabled: z.boolean().nullable().optional(),
+  auth: mcpAuth.optional(),
+})
+
+const toolEntry = z.strictObject({
+  type: z.literal('built_in').optional(),
+  enabled: z.literal(true).nullable().optional(),
+  permission,
```
asherdale: `permission` is optional for built-in tools, so requiring it here means valid entries like `ask_question: {}` can’t be shown in the builder. should we support omitted permissions?

frontend/apps/web/src/routes/AgentView.tsx
```diff
@@ -41,52 +61,110 @@ export function AgentView() {
...
+            <PageBreadcrumb
+              items={[
+                { label: activeOrg.name, to: '/' },
+                ...(project ? [{ label: project.name }] : []),
+                {
+                  label: 'Agents',
+                  to: '/projects/$projectId/agents' as const,
+                  params: { projectId },
+                },
+                ...(profile
```
asherdale: Image

these duplicate 'Basic Agent' breadcrumbs don't look great imo, and now this duplication would appear for every agent launched through the UI. my thought is that we could get rid of the profile breadcrumb, since there is already a link to the profile in the sidebar you've added... but maybe you can think of something better, i'd be good with whatever

asherdale (approved):
lgtm! just some straightforward suggestions

internal/storage/queries/agent_runtime.sql
cmsparks: yeah you're right

api/openapi/openapi.yaml
cmsparks: yeah i was kinda split on this, i'll add a patch endpoint instead

frontend/apps/web/src/components/agents/useAgentBuilderForm.ts
cmsparks: avoiding clobbering these keys

frontend/apps/web/src/components/agents/AgentSidebar.tsx
```diff
@@ -0,0 +1,163 @@
...
+      machine == null ? provisioningPollInterval : connectionStatePollInterval,
+  })
+  return (
+    <SidebarMenuItem className="flex items-center justify-between gap-2 py-1.5 text-sm">
+      <span className="truncate">{machine?.display_name ?? machineId}</span>
+      {machine && (
+        <Badge variant="outline" className="capitalize">
+          {machine.connection_state}
+        </Badge>
+      )}
```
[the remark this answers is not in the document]
cmsparks: this is expected.

PR #349, opened 2026-08-12
feat: add Amazon Bedrock model providers

asherdale:
- add Amazon Bedrock as a model provider using Bedrock API keys across Chat Completions, Responses, and Anthropic Messages
- add region and API format setup while keeping model configuration manual after a successful credential probe
- persist the Bedrock API variant through the database schema, OpenAPI contract, and generated SDK types


frontend/apps/web/src/components/org/CreateModelProviderDialog.tsx
```diff
@@ -69,12 +73,12 @@ export function CreateModelProviderDialog({
     const submissionGeneration = ++providerSubmissionGeneration.current
     setStatus(idle)
     try {
-      const result = await createModelProvider.mutateAsync({
-        name: values.name.trim(),
-        preset: values.preset,
-        credential_secret_id: values.secretId,
-      })
+      const result = await createModelProvider.mutateAsync(createModelProviderRequest(values))
```
cmsparks: tiny nit: i don't think we need the extra helper function, it's only used in one place, can we just inline createModelProviderRequest?

cmsparks (approved):
lgtm, one tiny nit

frontend/apps/web/src/components/org/CreateModelProviderDialog.tsx
asherdale: agreed, inlined this

PR #352, opened 2026-08-13
feat: stream tool call updates over SSE

asherdale:
- publish transaction-committed notifications for every tool-call lifecycle transition through the existing in-memory and Redis routing paths
- emit agent-scoped `tool_call_update` SSE frames containing only `tool_call_id` and `state`, independently of optional model-output deltas
- document the best-effort reconnect contract and update the OpenAPI schema, generated clients, and React stream handling


tools/ci/openapi-compat/approved-breaking-changes.txt
```diff
@@ -0,0 +1 @@
+OMN2-237 intentional SSE protocol addition: GET /api/v1/orgs/{orgID}/projects/{projectID}/agents/{agentID}/events/stream added `#/components/schemas/ToolCallUpdate` to the response body `oneOf` list for the response status `200`
```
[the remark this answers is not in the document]
asherdale: wrong, confirmed locally and in the CI for this PR: the check fails with this finding when the allowlist is omitted and passes with the committed entry

api/openapi/openapi.yaml
```diff
@@ -8109,6 +8110,16 @@ components:
...
           $ref: '#/components/schemas/Timestamp'
+    ToolCallUpdate:
+      description: An ephemeral notification that a tool call entered a lifecycle state.
+      type: object
+      additionalProperties: false
+      required: [tool_call_id, state]
+      properties:
+        tool_call_id:
+          $ref: '#/components/schemas/ToolCallID'
+        state:
```
cmsparks, line 8121: We don't have to make this change, this is just an idea

MCP tool calls can have progress updates: https://modelcontextprotocol.io/specification/2025-03-26/basic/utilities/progress

I don't think we have to build this functionality into the type now, but maybe it would make sense for the ToolCallUpdate to be more generic, like:

```typescript
{
  tool_call_id: "tcl_xyz",
  update: {
    type: "update_lifecycle",
    state: ToolCallState
  }
}
```

Then, we could add type: "progress" later to the same type without needing to create a separate ToolCallProgress type or making a breaking change to ToolCallUpdate.

cmd/api/main.go
```diff
@@ -371,6 +372,7 @@ func apiOptions(
 		httpapi.WithDaemonNotifications(redisBus, presence, replicaID),
 		httpapi.WithDaemonReplyPublisher(redisBus),
 		httpapi.WithAgentEventWakeupSubscriber(redisBus),
+		httpapi.WithAgentToolCallUpdateSubscriber(redisBus),
 		httpapi.WithAgentStreamDeltaSubscriber(redisBus),
```
cmsparks, line 376: it might make sense for us to combine these three subscribers at some point, because they all just send notifications to the same redis listener, but I dont think we need to worry about it right now

cmsparks (approved):
lgtm

api/openapi/openapi.yaml
asherdale: as discussed, we can leave as-is for now but easily extend the flat shape in the future to something like:
```
{
  "tool_call_id": "tcl_xyz",
  "kind": "progress",
  "progress": 50,
  "total": 100,
  "message": "Reticulating splines..."
}
```

PR #358, opened 2026-08-13
feat: Implement CRON schedules API

cmsparks:
* Add cron API endpoints
* Add cron tables
* Add cron worker handling


cmsparks:
made one small change to add claimed_until to resolve the cursor comment

migrations/000019_cron_triggers.sql
```diff
@@ -0,0 +1,47 @@
...
+    UNIQUE (project_id, idempotency_key)
+);
+
+CREATE UNIQUE INDEX cron_triggers_active_name_idx ON cron_triggers(project_id, name)
+    WHERE deleted_at IS NULL;
+
+CREATE INDEX cron_triggers_name_trgm_idx ON cron_triggers USING gin (name gin_trgm_ops)
+    WHERE deleted_at IS NULL;
+
+CREATE INDEX cron_triggers_due_idx ON cron_triggers(next_fire_after)
```
ksarangmath: the claim query orders by (next_fire_after, id), but this only covers next_fire_after, could we make this index match the query

internal/storage/queries/cron_triggers.sql
```diff
@@ -0,0 +1,186 @@
...
+  AND id = sqlc.arg(id)
+  AND deleted_at IS NULL;
+
+-- name: CompleteCronTriggerFiring :execrows
+UPDATE cron_triggers
+SET last_fired_at = transaction_timestamp(),
+    next_fire_after = sqlc.narg(next_fire_after),
+    claimed_until = NULL,
+    updated_at = statement_timestamp()
+WHERE project_id = sqlc.arg(project_id)
```
ksarangmath, line 181: This update should also check that the current operation's claim is active, because a different process could have taken the claim after an expired lease, and then the original process comes back to update these rows even though it doesn't have the claim. Might need to use some unique token to compare against.

internal/crontrigger/service.go
```diff
@@ -0,0 +1,213 @@
...
+			}
+		} else {
+			switch trigger.Target.Kind {
+			case executionstore.CronTriggerTargetAgentProfile:
+				stats.Launched++
+			case executionstore.CronTriggerTargetAgent:
+				stats.Inputs++
+			}
+		}
+		if err := s.execution.CompleteCronTriggerFiring(ctx, trigger.ProjectID, trigger.TriggerID); err != nil {
```
ksarangmath: this completes even on errors, but should we not update last_fired_at if it errors?

internal/crontrigger/service.go
```diff
@@ -0,0 +1,213 @@
...
+		return fmt.Errorf("load target agent profile: %w", err)
+	}
+	actor, err := cronTriggerActorParams(trigger)
+	if err != nil {
+		return err
+	}
+	launch, err := s.execution.LaunchAgent(ctx, executionstore.LaunchAgentInput{
+		ProjectID:     trigger.ProjectID,
+		ProfileID:     profile.ID,
+		AgentConfigID: profile.CurrentConfigID,
```
ksarangmath, line 150: nit: since we get profile above in a separate operation, this could not represent the actual current profile at this time. I'm fine with that cuz it's very small window and I don't think there's any correctness that relies on this, just pointing it out in case I missed something

internal/cronschedule/cronschedule.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/robfig/cron/v3"
+)
+
+var parser = cron.NewParser(
+	cron.Minute | cron.Hour | cron.Dom | cron.Month | cron.Dow,
+)
+
+func Validate(expression, timezone string) error {
+	if _, err := parser.Parse(expression); err != nil {
```
ksarangmath: robfig/cron accepts TZ= and CRON_TZ= prefixes, which might conflict with the other independent timezone that users can provide, should we reject those prefixes explicitly? or allow those and remove the explicit timezone (I lean towards the former, but idk what the standard is)

internal/cronschedule/template.go
```diff
@@ -0,0 +1,39 @@
...
+	return err
+}
+
+func RenderMessage(messageTemplate string, data map[string]any) (string, error) {
+	parsed, err := template.New("cron_message").Option("missingkey=error").Parse(messageTemplate)
+	if err != nil {
+		return "", fmt.Errorf("invalid message template: %w", err)
+	}
+	var rendered strings.Builder
+	if err := parsed.Execute(&rendered, data); err != nil {
```
ksarangmath: This executes user-controlled Go templates without output or execution limits, so a small template can consume excessive memory or CPU—for example, {{ printf "%1000000000s" "" }} requests roughly 1 GB of output. Since the only dynamic values are name, fired_at, and last_fired_at, could we replace the full template engine with fixed interpolation for those three placeholders?

ksarangmath: the template language is pretty expressive, if we do wanna keep it, we should probably limit it somehow

internal/crontrigger/service.go
```diff
@@ -0,0 +1,213 @@
...
+	return stats, nil
+}
+
+func (s *Service) fireTrigger(ctx context.Context, trigger executionstore.ClaimedCronTrigger) error {
+	message, err := cronschedule.RenderMessage(
+		trigger.MessageTemplate,
+		cronschedule.MessageData(trigger.Name, trigger.FiredAt, trigger.LastFiredAt),
+	)
+	if err != nil {
+		message = trigger.MessageTemplate
```
ksarangmath: if the rendering fails, should we just error instead of passing the raw template to the agent?

internal/storage/executionstore/agent_profiles_store.go
```diff
@@ -331,6 +331,11 @@ func (s *Store) DeleteAgentProfile(ctx context.Context, projectID, id ID) error
 	}); err != nil {
 		return fmt.Errorf("delete agent profile versions: %w", err)
 	}
+	if _, err := qtx.DeleteCronTriggersForAgentProfile(ctx, dbsqlc.DeleteCronTriggersForAgentProfileParams{
```
ksarangmath, line 430: also in project and org deletion, should we delete the cron triggers? I think right now they're just left

api/openapi/openapi.yaml
```diff
@@ -2630,6 +2630,222 @@ paths:
...
+        required: true
+        schema:
+          type: string
+          pattern: '^proj_[a-z2-7]{26}$'
+    post:
+      operationId: createCronTrigger
+      tags: [Configs and Profiles]
+      summary: Create cron trigger
+      description: >-
+        Creates a cron trigger that fires on a schedule. A trigger targeting an agent profile
```
ksarangmath, line 2726: should we mention that a missed trigger will be coalesced into the next trigger if a next trigger has been reached? e.g. if there's a cron that operates every minute, but our service goes down for 5 mins, the next trigger only triggers once, it doesn't try to retroactively run the previous 5 missed triggers. not sure what's standard cron behavior, if that's standard then we prob don't have to specify this

migrations/000019_cron_triggers.sql
```diff
@@ -0,0 +1,47 @@
...
+    project_id uuid NOT NULL,
+    name text NOT NULL,
+    agent_profile_id uuid,
+    agent_id uuid,
+    cron_expression text NOT NULL,
+    timezone text NOT NULL DEFAULT 'UTC',
+    message_template text NOT NULL,
+    enabled boolean NOT NULL DEFAULT true,
+    last_fired_at timestamptz,
+    next_fire_after timestamptz,
```
ksarangmath: nit: when enabled is true, should we have a constraint that ensure this is not null

internal/storage/executionstore/cron_triggers_store.go
```diff
@@ -0,0 +1,607 @@
...
+		RowLimit: int64(limit),
+	})
+	if err != nil {
+		return ClaimDueCronTriggersResult{}, fmt.Errorf("select due cron triggers: %w", err)
+	}
+	result := ClaimDueCronTriggersResult{}
+	claimedUntil := now.Add(cronTriggerClaimLease)
+	for _, row := range rows {
+		if _, err := cronschedule.Next(row.CronExpression, row.Timezone, now); err != nil {
+			if _, err := qtx.DisableCronTrigger(ctx, dbsqlc.DisableCronTriggerParams{
```
ksarangmath, line 531: nit: it might be opaque as to why the cron got disabled, i.e. the user can't see the reason for why their cron suddenly stopped working. we could add some disabled reason on the db, not sure if there's a better way

internal/crontrigger/service.go
```diff
@@ -0,0 +1,213 @@
...
+	displayName := trigger.Name
+	return &executionstore.ActorParams{
+		Provider:         executionstore.ActorProviderOmnara,
+		ProviderTenantID: tenantID,
+		ProviderUserID:   providerUserID,
+		DisplayName:      &displayName,
+	}, nil
+}
+
+func permanentFireError(err error) bool {
```
ksarangmath, line 220: Related to my previous comment, but right now triggers have no durable failure state. schedule-evaluation errors disable them without exposing the reason, while quota, capacity, or target-state errors leave them enabled and retrying after each lease. Can we expose the last failure or retry status so users can understand why a trigger is not firing? I'm also fine if this failure stuff is done in a followup, since it's purely additive.

ksarangmath (approved):
LGTM, design looks good, left some straightforward comments

internal/cronschedule/template.go
cmsparks: Going to add a limited writer, so that it fails to write if we hit some amount of bytes written, and a timeout via a goroutine. I think this should be largely comprehensive

api/openapi/openapi.yaml
cmsparks: Yeah, we can mention this in docs, I think that's expected though

internal/cronschedule/template.go
```diff
@@ -0,0 +1,156 @@
...
+	timer := time.NewTimer(renderTimeout)
+	defer timer.Stop()
+	select {
+	case err := <-execution:
+		if err != nil {
+			return "", fmt.Errorf("invalid message template: %w", err)
+		}
+	case <-timer.C:
+		return "", errors.New("invalid message template: rendering timed out")
+	}
```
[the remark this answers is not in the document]
cmsparks: this is expected because render message doesn't expose a context to cancel. technically this doesn't prevent certain cases of resource use, but at least discourages expensive templates.

PR #360, opened 2026-08-14
fix: clarify machine pool resource limits

asherdale:
- replace the ambiguous Resources summary with max total machine and memory capacity so pool limits are visible at a glance
- expand pool details with total, default, minimum, and maximum CPU and memory limits using explicit per-machine labels and related ordering
- omit unset optional limits while preserving explicit zero values, and show unavailable summary memory with an em dash
- update required vuln dependency so CI passes

Screenshot 1:
image

Screenshot 2:
image


ksarangmath (approved):
LGTM, couple of nits:
Instead of "Max total" in the UI, could we remove that? so it's just Machines and Memory? 

Also, could we throughout the UI use GB instead of MB for memory? this could be a new ticket, but via the UI it seems like people will only care about allocating by GB, and it looks nicer

PR #361, opened 2026-08-14
fix: retry machine provider provisioning

cmsparks:
## Summary
- retry idempotent machine provider provisioning up to three times before entering the durable reconciliation failure path
- use bounded backoff and honor provider `Retry-After` hints
- preserve provider resource observations across ambiguous failures and reject conflicting resource identities

## Tests
- `GOTOOLCHAIN=auto go test ./internal/machinepool ./internal/machinepool/providers/blaxel ./internal/machinepool/providers/daytona ./internal/machinepool/providers/unikraft`

### DIFF
--- a/internal/machinepool/manager.go
+++ b/internal/machinepool/manager.go
@@ -23,8 +23,15 @@ const (
 	providerInspectionTimeout           = 5 * time.Second
 	providerDeletionTimeout             = 5 * time.Second
 	providerFailureRetryDelay           = time.Minute
+	providerProvisionRetryAttempts      = 4
 	poolMachineReconcileConcurrency     = 8
 )
+
+var providerProvisionRetryDelays = [...]time.Duration{
+	time.Second,
+	5 * time.Second,
+	15 * time.Second,
+}
 
 type Manager struct {
 	Execution                  *executionstore.Store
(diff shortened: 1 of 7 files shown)


internal/machinepool/manager.go
```diff
@@ -210,6 +218,80 @@ func (m Manager) ProvisionMachine(ctx context.Context, orgID, machineID storage.
...
+			ctx,
+			installationID,
+			machineID,
+			machineProvisioning,
+			machineToken,
+			machineEnv,
+		)
+		provisionErr = err
+		if errors.Is(err, providers.ErrResourceReplaced) {
+			observed = providers.ProvisionMachineResult{}
```
asherdale: should replacement also update the stored `provider_resource_id`? this only clears the id held by this helper. for Daytona, if a previous provisioning attempt recorded the old Daytona sandbox id, the recreated sandbox gets a new id, but `RecordPoolMachineProvisioningResource` refuses to replace the old one, so the machine can't finish provisioning

internal/machinepool/providers/daytona/provider.go
```diff
@@ -197,8 +197,9 @@ func (p *provider) ProvisionMachine(
 				return result, err
 			}
 			return providers.ProvisionMachineResult{}, fmt.Errorf(
```
asherdale: does this need to wait for the deletion to finish before retrying? Daytona deletes are fire-and-forget, so the next call can conflict with the still-destroying sandbox and observe its old UUID again. once it disappears and is recreated, `provisionMachineWithRetry` rejects the new UUID as conflicting

internal/machinepool/providers/daytona/provider.go
```diff
@@ -197,8 +197,9 @@ func (p *provider) ProvisionMachine(
...
-				"daytona sandbox %q was deleted; provisioning must be retried",
+				"daytona sandbox %q was deleted; provisioning must be retried: %w",
 				name,
+				providers.ErrResourceReplaced,
```
asherdale: does the `!found` path below also need to wrap `ErrResourceReplaced`? Daytona deletion is fire-and-forget, so a retry can observe the old sandbox while it is still `destroying` and then hit `!found` once deletion finishes. without the replacement signal, the old UUID is preserved and the next create's new UUID is rejected as conflicting

internal/machinepool/manager.go
asherdale: should this also replace the stored `provider_resource_id`? clearing `observed` only affects retries within this call. for Daytona, if an earlier provisioning attempt stored the old Daytona UUID, this call can recreate the sandbox with a new UUID, but `RecordPoolMachineProvisioningResource` rejects it and the machine can't finish provisioning

internal/machinepool/providers/daytona/provider.go
asherdale: could a sandbox that disappears after being observed as `destroying` leave us holding its old UUID here, causing the replacement UUID to conflict on the next retry?

asherdale (approved):
lgtm!

internal/machinepool/providers/daytona/provider.go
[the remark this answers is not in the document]
cmsparks: low risk, if this happens we just don't retry

PR #362, opened 2026-08-14
feat: surface pending organization invitations

ksarangmath:
## Summary

- surface a pending-organization-invitation count in the sidebar and add an authenticated `/invitations` page that also works for users without an organization
- show each invitation’s organization name, assigned role, and copyable organization ID, with shared accept/decline UI across onboarding and the invitations page
- switch to the accepted organization before navigation and persist the active organization across refreshes
- link invitation emails directly to `/invitations`, including console-delivered local emails
- return invitation organization data consistently and harden invitation storage against stale or mismatched reads
- report existing-member and different-role pending invitations as normal business conflicts while allowing a declined or revoked invitation to be sent again
- keep member invitations on the Members page, use a standard role select, and close to the refreshed pending-invitation list after a successful send
- keep solid primary-indigo badge content white in every sidebar state
- cover invitation access, decline/re-invite behavior, email links, and conflict semantics with regression tests

image


frontend/apps/web/e2e/web.spec.ts
```diff
@@ -125,6 +126,37 @@ test('creates an organization from a project page', async ({ page }) => {
   expect(failures).toEqual([])
 })
 
+test('opens and declines pending invitations outside onboarding', async ({ page }) => {
```
asherdale: could we cover the accept path here and assert that the dashboard switches to the invited org? this test only covers decline, so accepting an invitation isn't covered

internal/storage/identitystore/org_invitations_store.go
```diff
@@ -306,29 +315,29 @@ func (s *Store) answerOrgInvitation(
...
 	if accept {
 		orgActive, err := qtx.OrgExistsActive(ctx, dbsqlc.OrgExistsActiveParams{ID: answered.OrgID})
 		if err != nil {
-			return OrgInvitationRecord{}, fmt.Errorf("check org for invitation accept: %w", err)
+			return OrgInvitationWithOrgNameRecord{}, fmt.Errorf("check org for invitation accept: %w", err)
 		}
 		if !orgActive {
 			// A pending invitation must never mint a membership in a deleted
 			// organization; deletion also revokes invitations, but the accept
 			// path guards against races.
```
asherdale: not added by this PR but could we delete this comment? it explains something very obvious in 3 lines

asherdale (approved):
lgtm 💪

frontend/apps/web/e2e/web.spec.ts
ksarangmath: Addressed in cf9a660d6: the same flow now declines one invitation, accepts another, then asserts navigation to `/` and that the invited organization is the active org in the dashboard switcher.

internal/storage/identitystore/org_invitations_store.go
ksarangmath: Removed in cf9a660d6.

frontend/apps/web/src/components/invitations/PendingInvitationList.tsx
```diff
@@ -0,0 +1,176 @@
...
+    }
+  }
+
+  async function copyOrganizationID(invitation: OrgInvitation) {
+    try {
+      await navigator.clipboard.writeText(invitation.org_id)
+      setCopiedInvitationID(invitation.id)
+    } catch {
+      setAction({ kind: 'error', message: 'Could not copy organization ID' })
+    }
```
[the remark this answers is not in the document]
ksarangmath: Also addressed in cf9a660d6: clipboard failures now use separate error state, so they cannot clear the in-flight accept/decline lock.

frontend/apps/web/src/components/active-org/ActiveOrgProvider.tsx
```diff
@@ -32,6 +32,7 @@ export function ActiveOrgProvider({ children }: { children: ReactNode }) {
     orgs,
     activeOrg,
     setActiveOrgId: (id) => {
+      writeLocal(STORAGE_KEY, id)
```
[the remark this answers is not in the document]
ksarangmath: Valid. Fixed structurally in c8f628532: persisted state now mirrors the explicitly selected org ID, never the derived fallback org, so a stale provider cannot overwrite a newly selected org during remount. I also moved organization creation to persist its selection before navigation, covering the same provider-boundary class of bug.

PR #367, opened 2026-08-16
refactor: enforce consistent resource name and slug policies

ksarangmath:
This PR makes a hard cutover to one resource-name policy while keeping machine-readable skill slugs separate.

- Organization, project, agent-profile, provider-config, configured-model, machine-pool, machine display, secret, API-key, token, cron-trigger, and device-flow client names now share a 64-code-point limit. Agent names use the same policy but may be empty to represent an unnamed agent.
- HTTP names must fit the limit as submitted. Accepted names are normalized to Unicode NFC and revalidated before storage and lookup; they are otherwise preserved rather than trimmed, while boundary whitespace, control or invisible characters, and invalid UTF-8 are rejected.
- Skill names remain lowercase ASCII slugs made of letters or numbers separated by single hyphens, with a 64-character limit. User profile display names are intentionally unchanged.
- Agent YAML and JSON resource references are normalized and validated before the existing name-to-ID resolution. Boundary whitespace is rejected instead of silently trimmed; authored source bytes remain unchanged.
- Migration 25 repairs existing relational names by trimming, normalizing, and truncating them, then installs versioned database constraints. Migration 26 removes the legacy top-level `name` from affected YAML and JSON configs, updates stored definitions and hashes, and fails on unexpected shapes or uniqueness collisions. It also removes the empty-source compatibility path and requires nonempty agent-config source.

There are no grandfathered invalid names, compatibility parsers, automatic deletions, or new ID-based YAML syntax.


internal/httpapi/secret_routes.go
```diff
@@ -479,6 +482,9 @@ func (s strictOpenAPIServer) UpdateSecret(
 		if *request.Body.Name == "" {
 			return nil, apierror.FromCode(openapi.ErrorCodeInvalidRequest, "name cannot be empty")
 		}
+		if err := resourcename.Validate("secret name", *request.Body.Name); err != nil {
```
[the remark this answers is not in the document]
ksarangmath: The three cited HTTP create paths already reach authoritative Go validation before persistence: secrets validate through secretstore, org API keys through prepareOrgAPIKeyInput, and model provider configs through createModelProviderConfigTx. Their InvalidRequest errors are also returned with field-specific detail, so adding route-level copies would duplicate the invariant without closing a bypass. The audit did find the deeper issue in this class: secretstore.CreateTx (used by internal cluster provisioning) bypassed the public preflight validator before reaching createSecretTx. I moved secret-name enforcement into that shared write sink and centralized create/update name validation, so internal transactional callers are covered even for Unicode format characters the PostgreSQL trigger cannot classify. I also added boundary tests for CreateTx and the two other cited store entry points. Fixed in eacd67e8b.

internal/httpapi/secret_routes.go
```diff
@@ -479,6 +482,9 @@ func (s strictOpenAPIServer) UpdateSecret(
...
+			return nil, apierror.FromCode(openapi.ErrorCodeInvalidRequest, err.Error())
+		}
```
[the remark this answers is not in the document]
ksarangmath: This is intentional at the HTTP boundary. Update name fields reference ResourceName, so the OpenAPI request validator rejects an invalid supplied name before these handlers, including overlength legacy values. ResourceName explicitly documents that unchanged grandfathered names must be omitted from updates, while responses use ResourceNameResponse so legacy values remain readable. The store and DB equality exceptions are still required because patch implementations carry the persisted name through when the request omits it, and because internal callers need the same grandfathering behavior. Removing only these handler checks would not create general pass-through: OpenAPI would still reject length violations while unsupported Unicode format characters would begin passing, producing an inconsistent contract. I audited the analogous org API key and configured-model update routes as well; they follow the same present-name validation and omitted-name pass-through model.

internal/storage/executionstore/agent_launch_store.go
```diff
@@ -49,6 +50,9 @@ func (s *Store) LaunchAgent(
 			"project, agent config, and launching principal are required",
 		)
 	}
+	if err := resourcename.Validate("agent name", input.Name); err != nil {
+		return LaunchAgentResult{}, storeerr.InvalidRequest(err)
+	}
```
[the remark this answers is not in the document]
ksarangmath: A profile-ID fallback would violate the hard-cutover policy: invalid stored profile names are migration failures, not values we rewrite. The actual gap was that the effective profile-derived launch name was not explicitly validated before insertion. This branch now validates it and returns InvalidRequest, with a regression test, while migration 22 validates every stored profile name before rollout. No fallback is added.

migrations/000025_resource_name_lengths.sql
```diff
@@ -0,0 +1,206 @@
...
+
+UPDATE agent_profiles
+SET name = resource_name_repair(name, 64)
+WHERE name IS DISTINCT FROM resource_name_repair(name, 64)
+  AND resource_name_is_valid(resource_name_repair(name, 64), false);
+
+UPDATE agents
+SET name = resource_name_repair(name, 64)
+WHERE name IS DISTINCT FROM resource_name_repair(name, 64)
+  AND resource_name_is_valid(resource_name_repair(name, 64), true);
```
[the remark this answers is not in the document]
ksarangmath: This is intentional, not a missed migration failure. Agent names are the one resource in this policy that use `DefaultableResourceName`: empty explicitly represents an unnamed/default state, and both `agents_name_policy` and migration 24 therefore call `resource_name_is_valid(name, true)`. Migration 23 is the requested repair cutover, so an invalid whitespace-only legacy value is trimmed to that valid canonical empty state. The migration integration test makes this contract explicit by seeding `agents.name = '  '` and asserting the repaired value is empty (`internal/dbmigrate/postgres_integration_test.go:167-220`). Making migration 24 reject that row would contradict the API/DB contract and the intended trim repair. No code change.

Makefile
```diff
@@ -165,15 +166,20 @@ state-migration-create:
...
 	done
 
 migration-check:
 	@for dir in $(MIGRATION_DIRS); do \
-		$(GOOSE) -env=none -dir "$$dir" validate || exit $$?; \
+		for migration_file in "$$dir"/*.sql "$$dir"/*.go; do \
+			test -e "$$migration_file" || continue; \
+			case "$$migration_file" in *_test.go) continue ;; esac; \
+			$(GOOSE) -env=none -dir "$$migration_file" validate || exit $$?; \
+		done; \
```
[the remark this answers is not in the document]
ksarangmath: I checked this against the pinned Goose v3.27.3 implementation and binary. This is not an empty validation: `validate` calls `gatherFilenames`, which stats the `-dir` value and, when it is a file, passes that file directly to `GatherStats`.

A direct probe of `migrations/000001_identity.sql` exited 0 and reported `sql / Txn=true / Up=75 / Down=0`. A direct probe of `000026_migrate_agent_config_names_test.go` exited 1 with `no init function`, proving individual files are parsed.

The per-file loop is necessary because the Goose validate directory glob includes `*_test.go`, while its runtime Go-migration collector skips those files. I added a concise Makefile comment in 1aa0cdb37 to document that upstream mismatch. No validation gap remains.

migrations/000026_migrate_agent_config_names.go
```diff
@@ -0,0 +1,315 @@
...
+	return config, nil
+}
+
+func migrateAgentConfigSource(
+	format string,
+	raw []byte,
+) ([]byte, bool, error) {
+	if format != "yaml" {
+		return nil, false, fmt.Errorf("unsupported legacy agent config source format %q", format)
+	}
```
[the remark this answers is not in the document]
ksarangmath: This is intentional fail-closed behavior in the one-time legacy-data migration, not a limitation of the application compiler. The production audit found no JSON row selected by `definition ? 'name' OR compiled_definition ? 'name'`; every affected legacy row uses the observed YAML shape handled here. We deliberately removed hypothetical legacy JSON rewriting to keep this migration scoped to the exact data being migrated.

Immediately before deployment, the production preflight will reassert that there are no affected JSON rows and no source-only legacy names. If that invariant changes, the migration should stop rather than silently rewrite an unaudited historical shape. No code change is needed for the current audited dataset.

tools/ci/migrationcheck/checker.go
```diff
@@ -306,14 +335,22 @@ func (snapshot gitSnapshot) listSQL(directory string) ([]string, error) {
...
-		if filePath != "" && path.Dir(filePath) == directory && path.Ext(filePath) == ".sql" {
+		if filePath != "" && path.Dir(filePath) == directory && isMigrationFile(filePath) {
 			files = append(files, filePath)
 		}
 	}
 	sort.Strings(files)
 	return files, nil
 }
 
+func isMigrationFile(filePath string) bool {
```
asherdale: this name feels inaccurate since the actual check of these files being in the migration directory is not included in this function... I'd prefer a name like `isGoOrSqlFile` to reduce potential misuse in the future

internal/agentconfig/source_schema.go
```diff
@@ -91,13 +93,21 @@ type AgentConfigMCPToolSource struct {
 }
 
 func ParseSource(format SourceFormat, raw []byte) (AgentConfigSource, error) {
+	return parseSource(format, raw)
```
asherdale: nit: can we remove the one-line `ParseSource`/`parseSource` split? their signatures are identical now, so `ParseSource` can own the implementation

internal/resourcename/resourcename.go
```diff
@@ -0,0 +1,76 @@
...
+	"fmt"
+	"unicode"
+	"unicode/utf8"
+
+	"golang.org/x/text/unicode/norm"
+)
+
+const MaxCodePoints = 64
+
+func CanonicalizeRequired(field, value string) (string, error) {
```
asherdale, line 13: nit: can we remove the presentation-only label plumbing here? `field`/`fieldLabel` is passed through the new Go functions in this file, dozens of callers, and the new frontend helpers solely to format error text. we should add that context where the error is returned or rendered instead, like we do for the rest of the repo

internal/storage/skillstore/store.go
```diff
@@ -13,6 +14,8 @@ import (
 	"github.com/omnara-ai/omnara/internal/storage/storeerr"
 )
 
+var ErrInvalidSkillName = errors.New("invalid skill name")
```
asherdale: should this and `ErrInvalidSecretName` live in `storeerr`? they’re both cross-layer sentinels interpreted by the HTTP layer, which is what `storeerr` centralizes

internal/httpapi/openapi_request_validator.go
```diff
@@ -61,16 +64,106 @@ func newOpenAPIRequestValidator() (middleware, error) {
...
+	if depth > maximumSchemaDepth {
+		return nil, false, errors.New("OpenAPI schema exceeds normalization depth")
+	}
+	schema := schemaRef.Value
+	if text, ok := value.(string); ok && schema.Extensions[openAPIUnicodeNormalizationExtension] == "NFC" {
+		normalized := norm.NFC.String(text)
+		return normalized, normalized != text, nil
+	}
+	changed := false
+	for _, nested := range schema.AllOf {
```
asherdale: should we add a contract check that normalized fields aren’t placed behind unsupported schema constructs like `oneOf`, `anyOf`, or `additionalProperties`? otherwise future schema changes will silently bypass normalization

internal/storage/executionstore/pool_machines_store.go
```diff
@@ -424,7 +423,7 @@ func listMachinePoolSources(
 	if err != nil {
 		return nil, err
 	}
-	configSource, err := agentconfig.ParseStoredSource(
+	configSource, err := agentconfig.ParseSource(
```
asherdale, line 426: does this strict reparse need an upgrade path for existing configs? previously this path used `ParseStoredSource` and `TrimSpace`, while `Compile` persisted the raw source after resolving trimmed references. `insertAgentConfigTx` protects new rows, but i couldn't find anything that validates or migrates existing rows, so a stored `machine_pool_name: "Build Pool "` will now fail here after upgrade

migrations/000025_resource_name_lengths.sql
```diff
@@ -0,0 +1,151 @@
...
+    ADD CONSTRAINT machine_pools_name_policy CHECK (resource_name_is_valid_v1(name));
+ALTER TABLE machines
+    ADD CONSTRAINT machines_display_name_policy CHECK (resource_name_is_valid_v1(display_name)),
+    ALTER COLUMN display_name DROP DEFAULT;
+ALTER TABLE machine_daemon_tokens
+    DROP CONSTRAINT machine_daemon_tokens_name_check,
+    ADD CONSTRAINT machine_daemon_tokens_name_policy CHECK (resource_name_is_valid_v1(name));
+ALTER TABLE cron_triggers
+    DROP CONSTRAINT cron_triggers_name_check,
+    ADD CONSTRAINT cron_triggers_name_policy CHECK (resource_name_is_valid_v1(name));
```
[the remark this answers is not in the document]
asherdale: before this PR, project names only had to be nonempty. an existing name over 64 characters makes migration 25 fail with `projects_name_policy ... violated by some row`, without identifying the row. should we migrate these rows or report the affected ids?

asherdale (approved):
lgtm, mostly straightforward nits

internal/resourcename/resourcename.go
ksarangmath: I cleaned up some drifted naming in the field param, but in for this pattern overall I think anything cleaner would require a bigger refactor of how we do error handling, which we could do later.

internal/storage/skillstore/store.go
ksarangmath: Ya, changed this

internal/httpapi/openapi_request_validator.go
ksarangmath: i just removed this stuff, it was doin too much

internal/storage/executionstore/pool_machines_store.go
ksarangmath: did a dry run on prod and everything was fine

migrations/000025_resource_name_lengths.sql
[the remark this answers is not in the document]
ksarangmath: did a dry run on prod and everything was fine, there's a lot of edge cases that could happen but we can do a hard cutover since there isn't any violating data

PR #370, opened 2026-08-17
fix: compact oversized model requests proactively

ksarangmath:
## The problem

Omnara could miss the point where a conversation needed compaction.

In the production incident, OpenRouter wrapped a real context-window overflow as a generic `502/provider_unavailable`. The message clearly said the input was too large, but Omnara trusted the wrapper, retried the same request repeatedly, and never compacted it.

The surrounding flow also had two correctness gaps: the configured context limit was not consistently enforced against the exact outgoing request, and provider-rejected opaque reasoning could reappear after compaction.

### Check the exact request before sending

Every supported route now serializes its final provider request and checks that exact body before transport. The configured context limit is authoritative, even when it is lower than the provider's real limit. Admission reserves the selected output allowance plus a safety margin.

Only the current prepared request can trigger proactive compaction. Historical provider usage remains useful for billing and diagnostics, but it cannot force a lossy checkpoint for a different request.

If the local estimate still undercounts, Omnara sends the request once. A deterministic context-window or payload rejection then enters compaction directly instead of retrying the unchanged request. Clear overflow language can override only generic gateway metadata; specific authentication, billing, permission, moderation, and rate-limit evidence remains authoritative.

### Compact without breaking conversation structure

Compaction summarizes a closed older prefix and keeps an exact recent tail. It preserves user turns and atomic tool-call/result groups, supports progressive checkpoints, and checks the reconstructed continuation before publishing the summary.

The summary output budget is chosen together with the source input budget:

- Prefer up to 16,384 summary tokens when there is room.
- Validate the configured model's normal output allowance against fixed provider constraints.
- If the preferred allowance conflicts with one of those constraints, fall back to the validated normal allowance.
- Use the smaller of the chosen allowance and the normal allowance as the summary-quality floor.
- Reduce the allowance only by the exact amount needed to fit.
- Stop before contacting the provider if even the smallest safe source cannot fit at that floor.

This avoids both false “irreducible” failures caused by reserving too much output and unsafe tiny summaries that would permanently replace useful history.

Compaction uses the runtime-resolved model and inherits its effective reasoning setting. It does not silently choose a cheaper reasoning level or switch providers. Compiled model overrides are now flattened in one place and shared by normal execution, compaction, storage validation, resolution, and API presentation.

### Keep rejected provider replay rejected

When a provider rejects signed or opaque reasoning/state, Omnara records the request's event frontier and retries once with provider-specific replay at or before that frontier removed. If the canonical request is rejected again at the same frontier, the call stops instead of looping.

The cutoff is durable and survives compaction. Older rejected replay stays suppressed, while new replay produced afterward remains eligible.

OpenRouter's unusual raw error envelopes are normalized only inside its adapter. Refinement is bounded to generic effective-400 failures; authoritative outer errors and masked 5xx responses remain untouched.

### Keep live-provider journeys symmetric

The shared live journeys now cover all four supported routes:

- OpenAI Responses
- OpenAI Chat Completions
- OpenRouter
- Anthropic Messages

Each route covers a normal service turn, Docker/process tools, compaction with checkpoint/continuation/recall, and the direct compaction runner. The provider-switching journey traverses all four routes, and the compaction journey verifies that useful tool-derived facts survive while disposable padding does not.

## Design boundary

Exact-request admission, provider-envelope decoding, shared error semantics, durable maintenance, and compaction execution remain separate. A future provider-native compaction strategy can be added behind the compaction boundary without replacing or entangling this client-side implementation.


cmsparks (approved):
lgtm

PR #371, opened 2026-08-17
feat: Implement Omnara CLI from openapi-based SDK

cmsparks:
Adds a declarative CLI for the Omnara API based on the openapi specification. The core idea is that we have a `manifest.ts` file, which defines every command and command group. Then, based on zod schemas and SDK functions, we can autogenerate the input format and the output formatting.

There are two additional commands here outside of just the CLI implementation:
* context, which gets/sets a persistent org and project id to use for commands
* whoami, which fetches auth and user info
* skills, special case for uploading archives / directories

This implementation doesn't include device flow login or grants (I have to figure out a nice ergonomic way to do grants)


frontend/packages/cli/src/factory.ts
```diff
@@ -0,0 +1,304 @@
...
+          let value = explicit ?? context.resolve(ctx)
+          if (value === undefined && canPromptInteractively()) {
+            value = await context.prompt(ctx.client, path)
+            if (!usedExplicitOverride) saveContextDefault(context.configKey, value)
+          }
+          if (value === undefined) {
+            throw new CliInputError(`missing ${context.optionKey}: ${context.describe}`)
+          }
+          path[context.key] = value
+        }
```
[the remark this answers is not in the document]
asherdale: i think this is still reproducible on the current head: `--org org-new` still resolves `projectID` from the saved default, so project-scoped commands can send `org-new` with a project from the old org

frontend/packages/cli/src/factory.ts
```diff
@@ -0,0 +1,272 @@
...
+        input.path = parseWithSchema(spec.path, path, 'arguments')
+      }
+      if (spec.query) {
+        const query = collectFlagValues(queryFlags, options)
+        input.query = parseWithSchema(spec.query, query, 'query flags')
+      }
+      if (spec.body) {
+        const base =
+          options.body === undefined ? {} : parseWithSchema(zBodyObject, options.body, '--body')
+        const body = { ...base, ...collectFlagValues(bodyFlags, options) }
```
asherdale: should we deep-merge field flags into `--body` here? `--material-value` currently replaces the entire `material` object, dropping fields like `kind`

asherdale: should we deep-merge field flags into `--body` here? `--material-value` currently replaces the entire `material` object, dropping fields like `kind`

asherdale (approved):
lgtm 💪

frontend/packages/cli/src/factory.ts
cmsparks: yeah added a deep merge functionality

PR #372, opened 2026-08-17
fix: include BYO machines in machine observation tools

asherdale:
- fixes OMN2-266 by replacing the pool-only read path behind `list_machines` and `inspect_machine` with unified observations over agent machine bindings
- includes attached BYO and pool-backed machines in tool results, with source, binding, display, and executable-state metadata
- preserves ref-based inspection of released pool bindings for deletion lifecycles while keeping released bindings out of machine lists
- updates tool contracts and docs, with coverage for BYO-only and mixed BYO plus pool selection flows

### DIFF
--- a/internal/harness/tools/machine_execution_target_integration_test.go
+++ b/internal/harness/tools/machine_execution_target_integration_test.go
@@ -192,6 +192,203 @@ func TestResolveMachineExecutionTargetUsesMachineRefSelection(t *testing.T) {
 	_ = secondAgentBinding
 }
 
+func TestMachineObservationToolsIncludeAttachedBYOMachine(t *testing.T) {
+	ctx := context.Background()
+	fixture := newMachineDispatchFixture(t, ctx, "observe-byo")
+	byo := createExecutableBinding(
+		t,
+		ctx,
+		fixture.Store,
+		fixture.UserID,
+		"observe-byo",
+		fixture.Now.Add(5*time.Second),
+	)
+	machineRef := "mchr-byo001"
+	if _, err := fixture.Pool.Exec(
+		ctx,
+		`INSERT INTO agent_machine_bindings(
+		   org_id, project_id, agent_id, machine_id, machine_ref, binding_kind, state,
+		   description, cwd, created_at, updated_at
+		 )
+		 VALUES ($1, $2, $3, $4, $5, 'explicit', 'attached', $6, $7, $8, $8)`,
+		toolsTestOrgID,
+		toolsTestProjectID,
+		fixture.Launch.Agent.ID,
+		byo.MachineID,
+		machineRef,
+		"local checkout",
+		"/checkout",
+		fixture.Now.Add(6*time.Second),
+	); err != nil {
+		t.Fatalf("attach BYO machine: %v", err)
+	}
+
+	listCall := model.ToolCall{ID: "call_observe-byo-list", Name: "list_machines", Input: json.RawMessage(`{}`)}
+	inspectCall := model.ToolCall{ID: "call_observe-byo-inspect", Name: "inspect_machine", Input: json.RawMessage(`{}`)}
+	mixedInspectCall := model.ToolCall{
+		ID:    "call_observe-byo-mixed-inspect",
+		Name:  "inspect_machine",
+		Input: json.RawMessage(`{}`),
+	}
+	createPoolCall := model.ToolCall{ID: "call_observe-byo-create-pool", Name: "create_machine", Input: json.RawMessage(`{}`)}
+	toolCalls, lock, admitted, contextRecord := recordMachineToolCallsForDirectStoreTest(
+		t,
+		ctx,
+		fixture.Store,
+		fixture.Launch.Agent.ID,
+		fixture.UserID,
+		fixture.Config.ID,
+		"observe-byo",
+		[]model.ToolCall{listCall, inspectCall, mixedInspectCall, createPoolCall},
+		fixture.Now.Add(7*time.Second),
+	)
+	for _, index := range []int{0, 3} {
+		if _, err := fixture.Store.Execution().MarkToolCallReady(
+			ctx,
+			executionstore.MarkToolCallReadyInput{
+				ProjectID:     toolsTestProjectID,
+				AgentID:       fixture.Launch.Agent.ID,
+				ID:            toolCalls[index].ID,
+				RuntimeLockID: lock.ID,
+			},
+		); err != nil {
+			t.Fatalf("allow %s: %v", toolCalls[index].ProviderCallID, err)
+		}
+	}
+	turn := Turn{
+		ProjectID:          toolsTestProjectID,
+		AgentID:            fixture.Launch.Agent.ID,
+		SourceEventID:      admitted.Events[0].ID,
+		RuntimeLockID:      lock.ID,
+		ModelCallContextID: contextRecord.ID,
+		Tools: map[string]ToolSpec{
+			"list_machines": {
+				Permission: toolpermission.DefaultSelection(toolpermission.ModeAlwaysAllow),
+			},
+			"inspect_machine": {
+				Permission: toolpermission.DefaultSelection(toolpermission.ModeAlwaysAsk),
+			},
+		},
+	}
+	executor := Executor{
+		Store: fixture.Store,
+		Now:   func() time.Time { return fixture.Now.Add(8 * time.Second) },
+	}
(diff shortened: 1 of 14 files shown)


internal/storage/queries/agent_machine_bindings.sql
```diff
@@ -272,6 +272,59 @@ WHERE binding.project_id = sqlc.arg(project_id)
...
+       machine.lifecycle_state,
+       machine.failure_report,
+       connection.connection_state,
+       coalesce(current_runtime.state_reason_code, '') AS connection_state_reason,
+       coalesce(machine.lifecycle_reason_code, '') AS lifecycle_reason_code,
+       machine.lifecycle_reason_message,
+       coalesce(pool.name, '') AS machine_pool_name,
+       coalesce((
+         binding.state = 'attached'
+         AND EXISTS (
```
ksarangmath, line 294: Could we move this constraint outside of the classification of "executable" and have it just be part of the filter for SelectAgentMachineObservations? Right now list_machines or inspect_machine can return results for machines that don't have a project grant, but we should probably filter out machines that don't have a grant

ksarangmath: or maybe we do wanna show the machines, but hide any info about them, basically just saying that theyre unavaialable or not granted? that might be the more correct thing to do, cuz the agent binding can exist, it's just that the project doesn't have a grant for the machine

internal/storage/queries/agent_machine_bindings.sql
```diff
@@ -272,6 +272,59 @@ WHERE binding.project_id = sqlc.arg(project_id)
...
+           SELECT 1
+           FROM project_machine_grants pmgrant
+           WHERE pmgrant.project_id = binding.project_id
+             AND pmgrant.machine_id = binding.machine_id
+         )
+         AND machine.deleted_at IS NULL
+         AND machine.lifecycle_state = 'active'
+         AND connection.connection_state IN ('online', 'asleep')
+       ), false)::boolean AS executable
```
ksarangmath, line 303: nit: this defines an executable predicate, but also above we have ListExecutableAgentMachineBindings. I haven't seen where they're used, but is there some redundancy here? Like it could point to some redundancy in what tool calls return, or just that we've implemented the same thing twice. Ideally we have the logic in one place so they don't go out of sync

internal/storage/executionstore/agent_machine_observations_store.go
```diff
@@ -0,0 +1,138 @@
...
+	LifecycleState         MachineLifecycleState    `json:"lifecycle_state"`
+	ConnectionState        MachineConnectionState   `json:"connection_state"`
+	ConnectionStateReason  string                   `json:"connection_state_reason,omitempty"`
+	Description            string                   `json:"description"`
+	Cwd                    string                   `json:"cwd"`
+	Executable             bool                     `json:"executable"`
+	LifecycleReasonCode    string                   `json:"lifecycle_reason_code"`
+	LifecycleReasonMessage string                   `json:"lifecycle_reason_message"`
+	FailureReport          json.RawMessage          `json:"failure_report,omitempty"`
+	CreatedAt              time.Time                `json:"created_at"`
```
ksarangmath, line 30: nit: These timestamps come from the binding, but the payload also contains machine and connection state that can change without updated_at changing. Could the names make it clear that these are binding timestamps?

internal/harness/tools/machine_tools.go
```diff
@@ -499,8 +503,8 @@ func machineObservation(record executionstore.PoolMachineRecord) machineObservat
 	}
 	return machineObservationPayload{
```
ksarangmath, line 504: This mapper is used by the create/delete results, but it doesnt populate source_kind, binding_kind, or display_name fields that list_machiesn inspect_machine return. Since those values are available on PoolMachineRecord, should the shared result shape be consistent?

internal/storage/executionstore/pool_machines_integration_test.go
ksarangmath, line 1695: nit, not caused by your PR but somewhat related: This manually releases the binding while the machine is still active and its project grant still exists. Production deletion releases the binding after the machine is deleted and removes the grant. Could this test model the real final state? Otherwise a grant-filtering change could pass here while breaking inspection of an actual deletion tombstone.

internal/storage/executionstore/agent_machine_observations_store.go
```diff
@@ -0,0 +1,138 @@
...
+	q *dbsqlc.Queries,
+	projectID, agentID ID,
+	machineRef string,
+) (AgentMachineObservationRecord, error) {
+	records, err := selectAgentMachineObservations(
+		ctx,
+		q,
+		projectID,
+		agentID,
+		&machineRef,
```
ksarangmath, line 84: a comment might be warranted here to explain that inspect machine can return a pool machine that has been released, i.e. no grant exists, but it was once granted to the agent/project

internal/harness/tools/machine_execution_target_integration_test.go
```diff
@@ -192,6 +192,203 @@ func TestResolveMachineExecutionTargetUsesMachineRefSelection(t *testing.T) {
...
+	byo := createExecutableBinding(
+		t,
+		ctx,
+		fixture.Store,
+		fixture.UserID,
+		"observe-byo",
+		fixture.Now.Add(5*time.Second),
+	)
+	machineRef := "mchr-byo001"
+	if _, err := fixture.Pool.Exec(
```
ksarangmath, line 207: nit: This inserts the BYO binding directly rather than creating it through the production launch/binding path, should we use the production path instead to test that in this integration test?

internal/harness/tools/machine_execution_target_integration_test.go
```diff
@@ -192,6 +192,203 @@ func TestResolveMachineExecutionTargetUsesMachineRefSelection(t *testing.T) {
...
+	}
+	if err := executor.PrepareToolCallPermission(ctx, turn, mixedInspectCall); err != nil {
+		t.Fatalf("prepare mixed-source inspect_machine permission: %v", err)
+	}
+	mixedResult, err := executor.Dispatch(ctx, turn, mixedInspectCall)
+	if err != nil {
+		t.Fatalf("dispatch mixed-source inspect_machine: %v", err)
+	}
+	mixedBody := toolResultMapFromTestParts(t, mixedResult.ContentParts)
+	if mixedBody["error_code"] != ErrMachineSelectionRequired.Error() ||
```
ksarangmath, line 369: nit, not related to your PR: When inspect_machine is called without a ref and multiple machines exist, the error code depends on the permission mode: one path returns machine_selection_required, while another returns inspect_machine_failed. Could both paths return the same error?

ksarangmath (approved):
LGTM, just a few comments

internal/storage/queries/agent_machine_bindings.sql
asherdale: went with the second suggestion: bindings without a current project grant remain visible, but the result now returns `project_grant_missing: true`, forces `executable: false`, and redacts the unavailable machine and runtime details

internal/storage/queries/agent_machine_bindings.sql
asherdale: good catch, both queries now use `machine_connection_states`. they stay separate because one filters rows while the other returns an `executable` flag

internal/storage/executionstore/agent_machine_observations_store.go
asherdale: renamed these to `BindingCreatedAt` and `BindingUpdatedAt`

internal/harness/tools/machine_tools.go
asherdale: agreed, the create/delete mapper now includes `source_kind`, `binding_kind`, and `display_name`

internal/storage/executionstore/pool_machines_integration_test.go
asherdale: updated the test to use the real deletion lifecycle, including grant removal and tombstone inspection

internal/harness/tools/machine_execution_target_integration_test.go
asherdale: replaced the direct insert with the production config and `LaunchAgent` binding path

internal/harness/tools/machine_execution_target_integration_test.go
asherdale: fixed, both permission modes now return the same `machine_selection_required` result

PR #374, opened 2026-08-18
feat: Add cron trigger UI

cmsparks:
Add cron trigger UI to frontend:
* in agent sidebar
* in agent profile tabs

### DIFF
--- a/frontend/apps/web/src/components/agents/CronTriggersSection.tsx
+++ b/frontend/apps/web/src/components/agents/CronTriggersSection.tsx
@@ -7,11 +7,20 @@ import {
 } from '@omnara/react'
 import { type CronTrigger, type CronTriggerTarget } from '@omnara/sdk'
 import { useForm } from '@tanstack/react-form'
+import cronstrue from 'cronstrue'
 import { Trash2 } from 'lucide-react'
 import { useState } from 'react'
 
 import { Badge } from '@/components/ui/badge'
 import { Button } from '@/components/ui/button'
+import {
+  Combobox,
+  ComboboxContent,
+  ComboboxEmpty,
+  ComboboxInput,
+  ComboboxItem,
+  ComboboxList,
+} from '@/components/ui/combobox'
 import {
   Dialog,
   DialogContent,
@@ -27,6 +36,21 @@ import { Textarea } from '@/components/ui/textarea'
 import { useInfiniteQueryItems } from '@/hooks/use-infinite-query-items'
 import { formatDateTime } from '@/lib/format'
 import { errorMessage } from '@/lib/submit-status'
+
+const nextFireFormatter = new Intl.DateTimeFormat(undefined, {
+  year: 'numeric',
+  month: 'short',
+  day: 'numeric',
+  hour: 'numeric',
+  minute: '2-digit',
+  timeZoneName: 'short',
+})
+
+function nextFireLabel(value: string) {
+  const date = new Date(value)
+  if (Number.isNaN(date.getTime())) return ''
+  return ` · next ${nextFireFormatter.format(date)}`
+}
 
 export function CronTriggersList({
   orgId,
@@ -60,8 +84,15 @@ export function CronTriggersList({
                 </div>
                 <p className="text-muted-foreground truncate text-xs">
                   <span className="font-mono">{trigger.cron}</span> · {trigger.timezone}
-                  {trigger.next_fire_at && ` · next ${formatDateTime(trigger.next_fire_at)}`}
+                  {trigger.next_fire_at && nextFireLabel(trigger.next_fire_at)}
                 </p>
+                {trigger.failure_report && (
+                  <p className="text-destructive break-words text-xs">
+                    Failed {formatDateTime(trigger.failure_report.failed_at)}:{' '}
+                    {trigger.failure_report.message} ·{' '}
+                    {trigger.failure_report.will_retry ? 'will retry' : 'will not retry'}
+                  </p>
+                )}
               </div>
               {canManage && (
                 <div className="flex shrink-0 items-center gap-1">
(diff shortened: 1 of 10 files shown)


frontend/apps/web/src/components/agents/CronTriggersSection.tsx
```diff
@@ -0,0 +1,319 @@
...
+                  </FieldDescription>
+                </Field>
+              )}
+            </form.Field>
+            <form.Field name="timezone">
+              {(field) => (
+                <Field>
+                  <FieldLabel htmlFor="cron-trigger-timezone">Timezone</FieldLabel>
+                  <Input
+                    id="cron-trigger-timezone"
```
ksarangmath: Could this be a searchable timezone picker instead of free text? Users currently need to know an exact IANA identifier. We can display friendly names, default to the browser’s timezone, and still submit the IANA value required by the backend. A next-run preview would also make the selected timezone and DST behavior clear. Mentioning because I didn't know what IANA was and I tried "PT" and it errored cuz that wasn't IANA.

frontend/apps/web/src/components/agents/CronTriggersSection.tsx
```diff
@@ -0,0 +1,319 @@
...
+                    id="cron-trigger-message"
+                    required
+                    rows={4}
+                    value={field.state.value}
+                    onChange={(event) => {
+                      field.handleChange(event.target.value)
+                    }}
+                  />
+                  <FieldDescription>
+                    Sent on each firing. Go template syntax with {'{{.trigger.name}}'},{' '}
```
ksarangmath, line 377: nit: there's a case where someone could submit {{ printf "%[2]*[1]s" "" 100000000 }} and it takes up 1gb of ram, even though it eventually fails, but the work is still done, is there a way we can limit the work of the template so someone doesn't spam this and take down a server

frontend/apps/web/src/components/agents/CronTriggersSection.tsx
```diff
@@ -0,0 +1,319 @@
...
+    <div className="flex flex-col gap-2">
+      {triggers.length > 0 ? (
+        <ul className="bg-background flex flex-col divide-y rounded-md border">
+          {triggers.map((trigger) => (
+            <li key={trigger.id} className="flex items-center justify-between gap-3 px-3 py-2">
+              <div className="flex min-w-0 flex-col gap-0.5">
+                <div className="flex min-w-0 items-center gap-2 text-sm">
+                  <span className="truncate font-medium">{trigger.name}</span>
+                  {!trigger.enabled && <Badge variant="secondary">Disabled</Badge>}
+                  {trigger.failure_report && <Badge variant="destructive">Failing</Badge>}
```
ksarangmath, line 83: could we display the failure reason/whether it will retry

frontend/apps/web/src/components/agents/CronTriggersSection.tsx
ksarangmath: also would be nice if there was a "current timezone" default that could be picked based on the browser/computer

frontend/apps/web/src/components/agents/CronTriggersSection.tsx
```diff
@@ -0,0 +1,319 @@
...
+  const triggers = useInfiniteQueryItems(query)
+  const updateTrigger = useUpdateCronTrigger(orgId, projectId)
+  const deleteTrigger = useDeleteCronTrigger(orgId, projectId)
+
+  return (
+    <div className="flex flex-col gap-2">
+      {triggers.length > 0 ? (
+        <ul className="bg-background flex flex-col divide-y rounded-md border">
+          {triggers.map((trigger) => (
+            <li key={trigger.id} className="flex items-center justify-between gap-3 px-3 py-2">
```
ksarangmath, line 78: ik the API allows for editing crons, but i don't think it's in the UI (just pointing out, we can add that later)

frontend/apps/web/src/components/agents/CronTriggersSection.tsx
```diff
@@ -0,0 +1,319 @@
...
+              <div className="flex min-w-0 flex-col gap-0.5">
+                <div className="flex min-w-0 items-center gap-2 text-sm">
+                  <span className="truncate font-medium">{trigger.name}</span>
+                  {!trigger.enabled && <Badge variant="secondary">Disabled</Badge>}
+                  {trigger.failure_report && <Badge variant="destructive">Failing</Badge>}
+                </div>
+                <p className="text-muted-foreground truncate text-xs">
+                  <span className="font-mono">{trigger.cron}</span> · {trigger.timezone}
+                  {trigger.next_fire_at && ` · next ${formatDateTime(trigger.next_fire_at)}`}
```
ksarangmath: can we also display the timezone, cuz the cron timezone might be different than the browser's

frontend/apps/web/src/components/agents/CronTriggersSection.tsx
```diff
@@ -0,0 +1,319 @@
...
+                    }}
+                  />
+                </Field>
+              )}
+            </form.Field>
+            <form.Field name="cron">
+              {(field) => (
+                <Field>
+                  <FieldLabel htmlFor="cron-trigger-cron">Cron expression</FieldLabel>
+                  <Input
```
ksarangmath, line 314: nit: it'd be nice if there was a user readable summary of when the cron would happen (similar to https://crontab.guru/), and if validation could happen realtime instead of only when creation is clicked

ksarangmath (approved):
LGTM, left a few minor comments

frontend/apps/web/src/components/agents/CronTriggersSection.tsx
cmsparks: I don't think it's possible for us to estimate this at the time when we submit on the frontend. In the previous PR, I added limits in the API which should prevent most cases resource exhausion issues on the backend, though it seems like it doesn't handle this case. I'll add some more handling for shadowing printf to prevent the indexed syntax

frontend/apps/web/src/components/agents/CronTriggersSection.tsx
cmsparks: Yeah, I think we can add later if needed

internal/cronschedule/template.go
```diff
@@ -51,38 +85,62 @@ func validatePrintfFormat(format string) error {
...
+			}
+			remaining -= bound
 			i = next
 		}
+		if i < len(format) && format[i] == '[' {
+			return errors.New("explicit argument indexes are not supported")
+		}
+		if i < len(format) {
+			i++
+		}
```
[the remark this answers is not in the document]
cmsparks: unreleased, this is fine

PR #375, opened 2026-08-18
feat: add configurable idle deletion for machine pools

asherdale:
- adds `delete_after_idle_minutes` policies at the machine pool, project grant, and agent machine-source levels, with the most specific configured value taking precedence
- defines idleness from machine lifecycle and process activity while protecting queued, starting, and running processes plus queued or accepted process actions
- refreshes the idle window for applied process actions and routes expired machines through the existing claimed deletion and retry lifecycle
- reconciles deletions in bounded maintenance batches and exposes the policy through the API, configuration, and pool, grant, and agent UI


internal/storage/queries/execution_resources.sql
```diff
@@ -736,6 +743,29 @@ WHERE machine.org_id = sqlc.arg(org_id)
...
+    delete_attempts = machine.delete_attempts + 1,
+    provider_runtime_mismatch_since = NULL,
+    wake_attempt_expires_at = NULL,
+    updated_at = statement_timestamp()
+FROM candidate
+WHERE machine.org_id = sqlc.arg(org_id)
+  AND machine.id = candidate.id
+RETURNING machine.id, machine.org_id, machine.machine_pool_id, machine.source_kind, machine.display_name, machine.description, machine.provider, machine.lifecycle_state, machine.provider_resource_id, machine.provider_provision_attempted_at, 'offline'::text AS connection_state, machine.last_observed_at, machine.cpu, machine.memory_mb, machine.cwd, machine.env, machine.secret_env, machine.provider_options, coalesce(machine.idempotency_key, '') AS idempotency_key, coalesce(machine.lifecycle_reason_code, '') AS lifecycle_reason_code, machine.lifecycle_reason_message, machine.next_reconcile_after, machine.provision_attempts, machine.delete_attempts, machine.metadata, machine.deleted_at, machine.created_at, machine.updated_at, machine.lifecycle_changed_at, machine.lifecycle_version, false AS can_finalize_missing_provider_resource;
```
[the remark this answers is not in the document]
asherdale: this is already serialized by the machine-row lock. process/action creation requires the same lock with `lifecycle_state = 'active'`, so either the claim sees the new work or the insert is rejected. verified both race orders locally

migrations/000023_idle_pool_machine_deletion.sql
```diff
@@ -0,0 +1,96 @@
...
+    ADD CONSTRAINT agent_machine_bindings_delete_after_idle_minutes_check
+        CHECK (delete_after_idle_minutes IS NULL OR delete_after_idle_minutes > 0);
+
+CREATE UNIQUE INDEX agent_machine_bindings_one_attached_pool_machine_idx
+    ON agent_machine_bindings(org_id, machine_id)
+    WHERE binding_kind = 'pool' AND state = 'attached';
+
+-- Tracks activity that should postpone pool-machine idle deletion, not every process row update.
+ALTER TABLE processes
+    ADD COLUMN last_activity_at timestamptz NOT NULL DEFAULT statement_timestamp();
```
ksarangmath, line 24: shoudl we have machines.last_activity_at, this would capture things like skill usage/invocation? idle feels like a property of the machine, not an individual process

migrations/000023_idle_pool_machine_deletion.sql
```diff
@@ -0,0 +1,96 @@
+-- +goose Up
+
+ALTER TABLE machine_pools
+    ADD COLUMN delete_after_idle_minutes integer,
+    ADD CONSTRAINT machine_pools_delete_after_idle_minutes_check
```
ksarangmath, line 5: an we make null mean inherit, 0 mean explicitly disabled, 1–4 invalid, and >=5 the actual timeout? the most specific value should still win, so an agent can turn deletion back on after a project disables it. also, the query can’t just check IS NOT NULL, because then 0 would mean delete immediately.

ksarangmath (commented):
some initial comments

migrations/000023_idle_pool_machine_deletion.sql
asherdale: as discussed, leaving this as-is for now

migrations/000023_idle_pool_machine_deletion.sql
asherdale: good call, made this change in the latest commit

internal/storage/executionstore/process_action_completion_store.go
```diff
@@ -430,6 +430,15 @@ func completeDaemonProcessActionTx(
 	if err != nil {
 		return DaemonProcessActionReportApplication{}, fmt.Errorf("complete process action: %w", err)
 	}
+	if state == ProcessActionStateApplied {
+		if err := qtx.TouchProcessActivity(ctx, dbsqlc.TouchProcessActivityParams{
```
ksarangmath: nit: Should we update last_activity_at whenever an accepted process action finishes, rather than only when it is applied? Right now a failed, unknown, canceled, or reconciled action stops blocking deletion without refreshing the timestamp. If the process activity is already older than the timeout, the machine can then be deleted immediately instead of waiting the full idle period after the action finishes.

migrations/000023_idle_pool_machine_deletion.sql
```diff
@@ -0,0 +1,96 @@
...
+    FROM processes process
+    WHERE process.org_id = machine.org_id
+        AND process.machine_id = machine.id
+    ORDER BY process.last_activity_at DESC
+    LIMIT 1
+) activity ON true
+WHERE machine.source_kind = 'pool'
+    AND machine.lifecycle_state = 'active'
+    AND machine.deleted_at IS NULL
+    AND machine.last_observed_at IS NOT NULL
```
ksarangmath, line 58: Should an active wake_attempt_expires_at exclude the machine here? A wake commits that lease before calling the provider, so there could be a race of trying to delete/wake a machine.

api/openapi/openapi.yaml
```diff
@@ -10237,6 +10237,12 @@ components:
           format: int32
           minimum: 1
           maximum: 2147483647
+        delete_after_idle_minutes:
+          description: Idle minutes before deleting a machine from this pool. Omit to leave the pool default unset; a grant or agent source may still override it.
```
ksarangmath, line 10277: nit: we might need to also update docs/machines/pools.mdx with a new deletion reason

cmd/maintenance/main.go
```diff
@@ -386,8 +387,23 @@ func runMachinePoolMaintenanceLoop(
 ) {
 	ticker := time.NewTicker(interval)
 	defer ticker.Stop()
+	var nextIdleReconcileAt time.Time
 	for {
 		runMachinePoolMaintenanceTick(ctx, log, machinePoolManager)
+		if now := time.Now(); !now.Before(nextIdleReconcileAt) {
+			candidateCount, err := machinePoolManager.ReconcileIdleDeletion(
```
ksarangmath: Could we give this call the same panic protection as runMachinePoolMaintenanceTick, maybe we can make a shared helper for this, so that a panic doesn't kill the entire maintenance job

internal/storage/queries/execution_resources.sql
```diff
@@ -736,6 +743,29 @@ WHERE machine.org_id = sqlc.arg(org_id)
...
+  SELECT machine_id AS id
+  FROM expired_idle_pool_machine_candidates
+  WHERE org_id = sqlc.arg(org_id)
+    AND machine_id = sqlc.arg(id)
+)
+UPDATE machines machine
+SET lifecycle_state = 'deleting',
+    lifecycle_changed_at = statement_timestamp(),
+    lifecycle_version = machine.lifecycle_version + 1,
+    lifecycle_reason_code = 'idle_timeout',
```
ksarangmath, line 771: nit: this gets overwritten at deletion I think, but not a big deal for now. I think lifecycle_reason_message also might get overwritten

ksarangmath (approved):
LGTM

internal/storage/queries/execution_resources.sql
asherdale: as discussed, leaving this for now, but in the future we could add a `deletion_reason_code` and `deletion_reason_message` if needed/wanted

migrations/000023_idle_pool_machine_deletion.sql
```diff
@@ -0,0 +1,100 @@
...
+        binding.delete_after_idle_minutes,
+        pool_grant.delete_after_idle_minutes,
+        pool.delete_after_idle_minutes
+    ))
+    AND greatest(machine.lifecycle_changed_at, coalesce(activity.last_activity_at, machine.lifecycle_changed_at))
+        <= statement_timestamp() - make_interval(mins => coalesce(
+            binding.delete_after_idle_minutes,
+            pool_grant.delete_after_idle_minutes,
+            pool.delete_after_idle_minutes
+        ))
```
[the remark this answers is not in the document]
asherdale: already handled: `CompleteProcess` and `CompleteDaemonObservedProcess` update `last_activity_at` atomically with the terminal state, and accepted-action cancellation also refreshes it. no action needed

api/openapi/openapi.yaml
asherdale: good call, added idle deletion to the pool deletion docs along with the activity and override behavior

cmd/maintenance/main.go
asherdale: nice catch, both maintenance paths now use the same panic-recovery helper

migrations/000023_idle_pool_machine_deletion.sql
asherdale: yeah you're absolutely right. machines with an active wake lease are now excluded, with active and expired lease coverage

internal/storage/executionstore/process_action_completion_store.go
asherdale: makes a lot of sense. now every accepted-action terminal path now refreshes process activity, including cancellation and reconciliation

PR #377, opened 2026-08-18
feat: handle machine storage exhaustion

asherdale:
There are two main parts to this PR.

Part one:
- makes rejected-preparation cleanup crash-safe by retaining cleanup-only process state until local artifacts and SQLite rows are removed, without blocking idle sleep

Part two:
- detects filesystem, quota, and SQLite exhaustion before or after process acceptance, stops the affected supervisor, and reports machine_storage_exhausted
- carries storage failures through server completion and daemon reconciliation so agents receive durable failure results, unresolved actions settle, and machines recover when space is freed

This PR requires deploying the server before publishing the new daemon version.

### DIFF
--- a/internal/daemonprotocol/protocol.go
+++ b/internal/daemonprotocol/protocol.go
@@ -474,6 +474,8 @@ const (
 )
 
 const (
+	ProcessReasonMachineStorageExhausted      = "machine_storage_exhausted"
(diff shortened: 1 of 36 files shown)


cmsparks (approved):
lgtm!

PR #378, opened 2026-08-18
feat: Implement device flow login for CLI

cmsparks:


frontend/packages/cli/src/login.ts
```diff
@@ -0,0 +1,212 @@
...
+export function registerLoginCommand(program: Command, cli: CliConfig): void {
+  program
+    .command('login')
+    .description('Log in via the browser and save an API token')
+    .option('--no-browser', 'print the approval URL instead of opening a browser')
+    .option('--token-name <name>', 'name for the created API token')
+    .action(async (options: LoginOptions) => {
+      await runCliAction(async () => {
+        const report = canPromptInteractively() ? interactiveReporter(cli.baseUrl) : plainReporter()
+        const start = await startDeviceAuth({
```
asherdale, line 142: should we check that the existing config is valid before starting the device flow? `loadConfig()` ignores an unreadable config, so the user can complete approval before `updateConfigFile()` fails to save the new token. they then have to fix the config and repeat the flow, with the first PAT left active but unavailable

asherdale (approved):
lgtm

frontend/packages/cli/src/login.ts
cmsparks: yeah i'll add a handler for that case

PR #379, opened 2026-08-18
docs: Agent prompt for setting up Omnara

ishaansehgal99:
Prompt for agents to setup omnara


asherdale (approved):
worked well on my end

PR #380, opened 2026-08-18
feat: auto-add machine tools for machine sources

asherdale:
- Automatically add missing machine execution tools when a machine or pool source is selected in the basic agent builder while preserving existing tool permissions and avoiding duplicates.
- Show a focused warning with a one-click repair when a machine-backed config is missing recommended tools.
- Return non-blocking config warnings through config, profile, and launch API responses so API and CLI workflows expose incomplete machine-tool setups without mutating requests.
- Keep machine-backed documentation and examples aligned with the recommended tool set.


api/openapi/openapi.yaml
```diff
@@ -7871,6 +7871,13 @@ components:
           $ref: '#/components/schemas/AgentConfigModel'
         instruction_hash:
           type: string
+        warnings:
```
cmsparks, line 7889: can we type these warnings like we do with errors instead of just giving a string. i.e. it can be:

`type WarningMessages = { code: WarningCode, message: string }` where WarningCode is an enum with all possible codes

cmsparks (approved):
lgtm, pretty minor nit

api/openapi/openapi.yaml
asherdale: good call, updated this to match the existing `Error` pattern. `warnings` is now `Warning[]`, with each warning containing an enum-backed `code` and a human-readable `message`

PR #382, opened 2026-08-18
feat: Add interactive MCP and integration creation flows to CLI

cmsparks:
Add integration flows to add MCP and slack integration flows:
* Add `secret mcp-add` to create mcp secret
* Add `profile slack` to attach slack integration
* Add `profile mcp-add` to create mcp secret and update profile


frontend/packages/cli/package.json
```diff
@@ -1,25 +1,34 @@
 {
-  "name": "@omnara/cli",
-  "version": "0.0.0",
-  "private": true,
+  "name": "omnara",
+  "version": "1.0.0-beta.1",
```
asherdale, line 3: looks like `omnara@1.0.0-beta.1` is already published on npm, should we bump this so we can properly update it?

also seems like the built CLI crashes on Node 20 with `Dynamic require of "process" is not supported` because `yaml` is bundled into the ESM output. should we fix the bundle and add a startup smoke test?

frontend/packages/cli/src/browser.ts
```diff
@@ -0,0 +1,28 @@
...
+  args: string[]
+  windowsVerbatimArguments: boolean
+} {
+  if (process.platform === 'darwin') {
+    return { command: 'open', args: [url], windowsVerbatimArguments: false }
+  }
+  if (process.platform === 'win32') {
+    return {
+      command: 'cmd',
+      args: ['/c', 'start', '""', `"${url}"`],
```
asherdale: should `cmd.exe` corrupt percent-encoded OAuth URLs here? it expands `%...%` even inside quotes, so values like `https%3A%2F%2F...` may be altered before `start` receives them

asherdale (approved):
lgtm!

frontend/packages/cli/package.json
cmsparks: I'll update the version in the next PR when I setup the CI job for publishing (it probably should just be parametrized or something)

Node 20 is EOL and not supported anymore, but I can mark the supported node versions on publish. it does crash on 22/24 in that case though, and i fixed this

PR #383, opened 2026-08-18
feat: Add grants functionality for the CLI

cmsparks:
Add:
* grant skills list, add, delete
* grant secrets list, add, delete
* grant machines list, add, delete
* grant pools list, get, add, edit, delete
* grant models list, get, add, edit, delete

### DIFF
--- a/frontend/packages/cli/src/manifest.ts
+++ b/frontend/packages/cli/src/manifest.ts
@@ -645,6 +645,218 @@ export const commandGroups: CommandGroup[] = [
       }),
     ],
   },
+  {
+    name: 'grant',
+    aliases: ['grants'],
+    summary: 'Manage resource grants to projects',
+    groups: [
+      {
+        name: 'skills',
+        aliases: ['skill'],
+        summary: 'Manage skill grants',
+        operations: [
+          op({
+            verb: 'list',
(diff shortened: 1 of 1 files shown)


frontend/packages/cli/src/manifest.ts
```diff
@@ -645,6 +645,218 @@ export const commandGroups: CommandGroup[] = [
...
+            verb: 'delete',
+            summary: 'Remove a model grant',
+            fn: sdk.deleteProjectModelGrant,
+            format: formatVoid('deleted'),
+            path: schemas.zDeleteProjectModelGrantPath,
+          }),
+        ],
+      },
+    ],
+  },
```
asherdale, line 859: just noting that the PR description includes `grant models get`, but the CLI/API/SDK only exposes list/create/update/delete for model grants right now - not sure if you were meaning to add that now or later. could be an easy follow-up for later ofc

asherdale (approved):
lgtm!

frontend/packages/cli/src/manifest.ts
cmsparks: ah, yea i copy pasted something wrong, youre correct

PR #386, opened 2026-08-19
feat: improve machine pool management

asherdale:
- expose active machine, CPU, and memory usage in the machine-pool table and expand pool details
- share provider-aware form state and fields between machine-pool creation and editing
- allow restricted cluster-managed pools editing
- preserve those organization-owned cluster-pool settings during default reconciliation


frontend/apps/web/src/components/org/MachinePoolAdvancedSection.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+import {
+  derivedMemoryTotalCapPlaceholder,
+  derivedTotalCapPlaceholder,
+  type MachinePoolFormValues,
+} from './MachinePoolDialogState'
+import { MachinePoolInputField } from './MachinePoolInputField'
+import { machinePoolProviderDefinitions } from './machinePoolProviders'
+
+export function MachinePoolAdvancedSection({
```
cmsparks, line 16: Alignment in the advanced section looks a little weird. Mind adding a bit of padding above the first checkbox. Also  the limits titles end up on 2 lines in some cases.

also, i know this is preexisting, but i don't think we need most of these subtitles

Image

cmsparks (approved):
lgtm aside from some small ui nits

frontend/apps/web/src/components/org/MachinePoolAdvancedSection.tsx
asherdale: you're right on all of these - made these updates in the last commit (only left the subtitle for `runtime_protection` because its less self-explanatory)

image

frontend/apps/web/src/components/org/MachinePoolDialogState.ts
```diff
@@ -0,0 +1,511 @@
...
+    provider,
+    workspace: stringValue(pool.provider_config.workspace),
+    image: stringValue(options[definition.resource.key]),
+    location: stringValue(options[definition.location.key]),
+    startupScript: stringValue(options.startup_script),
+    cwd: pool.default_cwd,
+    envRows: envRowsFromRecord(pool.default_machine_env),
+    secretEnvRows: secretEnvRowsFromRecord(pool.default_machine_secret_env),
+    cpu: numberDraft(cpuValue) || machinePoolFormDefaults.cpu,
+    memoryGb: memoryGbDraft(memoryValue) || machinePoolFormDefaults.memoryGb,
```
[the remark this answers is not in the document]
asherdale: not reachable for a valid pool—provider validation requires these values. no change needed

PR #387, opened 2026-08-19
chore: remove active process and machine state from model context

asherdale:
- removes automatic active-process and attached-machine inventories from model prompts so rapidly changing runtime state no longer disrupts prompt-cache reuse
- deletes the now-unused model-context construction, normalization, provider serialization, and direct storage paths for that execution-state bundle
- preserves on-demand process and machine discovery through the existing list tools
- leaves machine-pool and integration-target prompt context unchanged

### DIFF
--- a/internal/storage/executionstore/process_daemon_completion_store.go
+++ b/internal/storage/executionstore/process_daemon_completion_store.go
@@ -502,21 +502,21 @@ func (s *Store) GetProcessByToolCall(
 	return getProcessByToolCallTx(ctx, s.pool, projectID, agentID, toolCallID)
 }
 
-func (r *ToolCallReader) ListActiveProcessesForContext(
+func (r *ToolCallReader) ListActiveProcesses(
 	ctx context.Context,
 ) ([]ActiveProcessRecord, error) {
 	t := r.transaction
-	return listActiveProcessesForContext(ctx, t.q, t.input.ProjectID, t.input.AgentID)
+	return listActiveProcesses(ctx, t.q, t.input.ProjectID, t.input.AgentID)
 }
 
-func listActiveProcessesForContext(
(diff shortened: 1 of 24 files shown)


internal/storage/executionstore/process_daemon_completion_store.go
```diff
@@ -502,16 +502,6 @@ func (s *Store) GetProcessByToolCall(
...
-	ctx context.Context,
-	projectID, agentID ID,
-) ([]ActiveProcessRecord, error) {
-	if isNilID(projectID) || isNilID(agentID) {
-		return nil, errors.New("project and agent are required")
-	}
-	return listActiveProcessesForContext(ctx, s.q, projectID, agentID)
-}
-
 func (r *ToolCallReader) ListActiveProcessesForContext(
```
ksarangmath: nit: this only gets called for the list_processes tool now, so we could remove the ForContext suffix from here and anywhere else that applies

ksarangmath (approved):
LGTM, might be worth double checking the tool call descriptions and error messages so that they point to the relevant tools to retrieve information that was removed in this PR (e.g. run command error on missing machine says use list machines to find machine ids, description of inspect machine says to use list machines to find machine ids to use here, etc). Cuz the model may have been relying on these context clues before, and now that we've removed it, it might perform worse because we haven't tuned the return values/descriptions/error messages

internal/storage/executionstore/process_daemon_completion_store.go
asherdale: renamed `ListActiveProcessesForContext`, updated tool descriptions to identify the refs returned by `list_machines`/`list_processes`, and pointed relevant schemas and errors to those tools

PR #389, opened 2026-08-19
chore: release flows for CLI and sdk

cmsparks:
Releasing the SDK and CLI in NPM

### DIFF
--- a/.github/workflows/sdk-release.yaml
+++ b/.github/workflows/sdk-release.yaml
@@ -1,0 +1,56 @@
+name: Release SDK
+
+on:
+  push:
+    tags:
+      - 'sdk-v[0-9]+.[0-9]+.[0-9]+'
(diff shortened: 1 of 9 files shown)


.github/workflows/sdk-release.yaml
```diff
@@ -0,0 +1,56 @@
...
+    runs-on: ubuntu-24.04
+    timeout-minutes: 15
+    permissions:
+      contents: read
+      id-token: write
+    defaults:
+      run:
+        working-directory: frontend
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```
asherdale, line 31: could we validate that `HEAD` belongs to `main` for these workflows, like we do for `cluster-release` and `omnarad-release`?

asherdale (approved):
LGTM

PR #390, opened 2026-08-19
feat: streamline agent creation and configuration

asherdale:
- simplify the agent creation form with aligned required fields, compact optional capabilities, and consistent resource controls
- default new agents to web search and fetch, move model grants into the selector, and hide implicit skill and integration tools from the builder
- remove first-message setup

image


cmsparks (approved):
lgtm

PR #391, opened 2026-08-20
feat: notify Slack users when agent launch fails

asherdale:
- notify Slack users when synchronous agent launches fail for known state, configuration, or capacity conflicts
- preserve existing handling for pre-launch conflicts and unexpected launch errors so unrelated failures do not generate misleading replies
- send the fallback through a metadata-free Slack message path and cover delivery plus transaction rollback with unit and database-backed integration tests

### DIFF
--- a/internal/integration/service.go
+++ b/internal/integration/service.go
@@ -17,8 +17,6 @@ type Service struct {
 	integrations integrationStore
 }
 
-var ErrAgentLaunchFailed = errors.New("launch agent for integration target")
-
 type executionStore interface {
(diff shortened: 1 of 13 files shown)


cmsparks (approved):
lgtm

internal/integration/service.go
```diff
@@ -17,6 +17,8 @@ type Service struct {
 	integrations integrationStore
 }
 
+var ErrAgentLaunchFailed = errors.New("launch agent for integration target")
```
cmsparks: can we move this into one of the common errors files? maybe in store or something

asherdale: good call, moved in the latest commit

PR #392, opened 2026-08-20
fix: hide retryable model attempt errors from agent chat

asherdale:
- fixes [OMN2-268](https://linear.app/omnara/issue/OMN2-268/hide-retryable-model-attempt-errors-from-the-agent-chat-ui) by treating model-output error deltas as failed attempt boundaries instead of chat-level errors
- removes partial output from a failed attempt while allowing a later backend retry to render normally
- keeps durable terminal errors and non-retryable stream failures visible while retryable service-unavailable interruptions reconnect silently
- extends the existing chat-stream coverage for recovered model attempts and silent stream reconnection


ksarangmath (approved):
LGTM

PR #401, opened 2026-08-20
feat: Add config attachments to agents

cmsparks:
Allow referencing yaml and json config files in agent upload and profile upload. also allow updating profiles and agent configs

### DIFF
--- a/frontend/packages/cli/src/config-attachment.ts
+++ b/frontend/packages/cli/src/config-attachment.ts
@@ -6,14 +6,20 @@ import * as z from 'zod'
 
 import { CliInputError } from './output.ts'
 
-export const zConfigAttachment = z.object({
-  config: zAgentConfigId.optional().describe('existing agent config ID'),
+export const zConfigSourceAttachment = z.object({
   file: z
     .string()
     .min(1)
     .optional()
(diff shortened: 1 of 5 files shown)


frontend/packages/cli/src/config-attachment.ts
```diff
@@ -0,0 +1,113 @@
...
+  return data.id
+}
+
+export async function resolveConfigSource(
+  client: OmnaraClient,
+  path: Record<string, unknown>,
+  attachment: ConfigAttachment,
+): Promise<ConfigSource> {
+  requireExactlyOne(attachment)
+  if (attachment.config === undefined) return renderConfigAttachment(attachment)
```
asherdale: should `agents update --config <id>` apply that exact config? this recompiles its source, so the result can differ from the ID the user passed

asherdale (approved):
lgtm 👍

besides the one nit below, i'd just remove the stray `agent.yaml` that bugbot commented on before merging

frontend/packages/cli/src/config-attachment.ts
cmsparks: Actually just removed this because it's not really expressible through the api, we can add it later.

PR #403, opened 2026-08-21
feat: add steering queue controls

asherdale:
- show queued steering messages with their actual content while an agent is working
- let users send queued messages immediately, remove them, or drag to reorder when multiple messages are waiting
- keep the queue synchronized with agent input events and load queued content blocks with a single batched query

### DIFF
--- a/frontend/apps/web/src/routes/AgentView.tsx
+++ b/frontend/apps/web/src/routes/AgentView.tsx
@@ -1,7 +1,6 @@
 import {
   useAgent,
   useAgentChat,
-  useAgentInputBacklog,
   useAgentInteractions,
   useAgentProfileQuery,
   useCancelAgent,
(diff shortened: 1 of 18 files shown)


frontend/packages/react/src/domains/agent-chat.ts
```diff
@@ -172,7 +173,10 @@ export class AgentChatSession {
         },
       })
       this.lastFailedSend = null
-      if (this.inputEchoLoaded(id)) this.clearPendingInput(id)
+      this.clearPendingInput(id)
+      void this.queryClient.invalidateQueries({
+        queryKey: agentInputBacklogQueryKey(this.client, this.scope),
+      })
```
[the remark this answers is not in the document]
asherdale: not an issue. `isWorking` does not control the SSE subscription, so the durable event still arrives, and the backlog is invalidated immediately after the POST

frontend/apps/web/src/routes/AgentView.tsx
```diff
@@ -68,6 +76,34 @@ export function AgentView() {
     return resolveInteraction.mutateAsync({ interactionID, body })
   }
 
+  async function sendQueuedInputNow(inputID: string) {
```
cmsparks: we're drilling these send props down through 3 components. generally, i think it's a code smell if we're creating onAction functions and passing application specific functions into components that aren't generic.

I think we can just do useAgentInputBacklog inside of the component itself and directly call the send now and remove now mutation function inside of the SortableQueueRow. moveQueuedInput can be just contained within the AgentInputQueue

cmsparks (approved):
one comment on prop drilling, otherwise lgtm

frontend/apps/web/src/routes/AgentView.tsx
asherdale: good call, moved the backlog hook and queue actions into `AgentInputQueue`

PR #405, opened 2026-08-21
feat: Add Agent CLI interactions

cmsparks:
* Add events
* add interactions
* add interactive cli chat


frontend/packages/cli/src/chat-ui.tsx
```diff
@@ -0,0 +1,422 @@
...
+  )
+}
+
+type TranscriptItem = { kind: 'older' } | { kind: 'message'; message: OmnaraUIMessage }
+
+function splitLive(
+  messages: OmnaraUIMessage[],
+  isWorking: boolean,
+): [OmnaraUIMessage[], OmnaraUIMessage[]] {
+  if (!isWorking) return [messages, []]
```
asherdale: should we keep local/submitted messages live until the durable event arrives (even if the agent is idle)?

frontend/packages/cli/src/chat-ui.tsx
```diff
@@ -0,0 +1,422 @@
...
+  const activity =
+    interaction == null ? currentActivity(chat.status, chat.isWorking, live) : undefined
+  const timer = useWorkTimer(interaction != null ? 'paused' : chat.isWorking ? 'working' : 'idle')
+  const lastDuration = interaction == null && !chat.isWorking ? timer.lastDuration : undefined
+
+  return (
+    <Box flexDirection="column">
+      <Transcript items={transcript} live={live} />
+      <BacklogView inputs={chat.inputBacklog.inputs} />
+      {chat.historyStatus === 'pending' && <Text dimColor>loading history…</Text>}
```
asherdale, line 395: in the rare case that `historyStatus` becomes `error`, should we render an error? right now the user wouldn't see anything

frontend/packages/cli/src/chat-ui.tsx
```diff
@@ -0,0 +1,422 @@
...
+      </Static>
+      {live.map((message) => (
+        <MessageView key={message.id} message={message} live />
+      ))}
+    </>
+  )
+}
+
+export function Chat({ scope }: { scope: AgentChatScope }) {
+  const chat = useAgentChat(scope)
```
asherdale: `createAgentChatInput` always prepends a hidden block saying “This message came from the Omnara web app” - so CLI sends are mislabeled. can we make that source/client-specific?

frontend/packages/cli/src/chat-ui.tsx
```diff
@@ -0,0 +1,422 @@
...
+      )
+    default:
+      return null
+  }
+}
+
+function MessageView({ message, live }: { message: OmnaraUIMessage; live: boolean }) {
+  if (message.role === 'user') {
+    const lines = message.parts.flatMap((part) =>
+      part.type === 'text' && part.text.trim() !== '' ? [part.text] : [],
```
asherdale, line 125: this branch keeps only text parts for user messages, so attachment-only messages disappear entirely. can we show an attachment marker here?

asherdale (approved):
lgtm 👌

frontend/packages/cli/src/chat-ui.tsx
cmsparks: I'll do that, also going to remove this text block if there's no send_integration_message tool

PR #406, opened 2026-08-21
test: make live compaction recall prompt benign

ksarangmath:
## Problem

The shared live compaction test described its retained value as an “opaque tool result” and told the model not to mention it before later asking for it. GPT-5.6 Luna and GLM-5.2 sometimes interpreted that wording as hidden or protected data and refused to answer, even though the checkpoint contained the value.

This made live CI fail after compaction had completed successfully.

## What changed

- Give the unpredictable tool result a benign generated sample identifier.
- Remove the secrecy-loaded and conflicting instructions.
- Preserve the same end-to-end contract: create a checkpoint, retain the exact tool-generated value, remove disposable padding, continue after compaction, and recall the value later.

The shared journey passes live across OpenAI Responses, OpenAI Chat Completions, OpenRouter, and Anthropic.


asherdale (approved):
LGTM

PR #407, opened 2026-08-21
feat: support home-relative machine paths

asherdale:
- Preserve home-relative working directories through machine, binding, and command-level path resolution.
- Expand ~ and ~/ paths on the machine daemon using its home environment, with failures reported through the existing terminal start-failure flow.
- Cover path composition, canonicalization, and missing-home behavior.


ksarangmath (approved):
LGTM

PR #408, opened 2026-08-22
feat: Add daemon creation functionality to CLI

cmsparks:
* update machines create to return the daemon token correctly
* add `create-local` which install the CLI and generates the daemon token


frontend/packages/cli/src/machine-setup.ts
```diff
@@ -0,0 +1,115 @@
...
+): Promise<MachineWithToken> {
+  const { client, path, body, report } = context
+  const { token_name, ...machineBody } = body
+  report.start('Creating machine')
+  const { data: machine } = await sdk.createMachine({ client, path, body: machineBody })
+  report.stop(`Machine created: ${machine.id}`)
+  report.start('Creating machine daemon token')
+  let created
+  try {
+    ;({ data: created } = await sdk.createByoMachineDaemonToken({
```
ksarangmath: Could we generalize the existing ConnectBYOMachine endpoint and use it here instead of making two server calls? It already creates the machine, daemon token, and project grants in one transaction. The frontend also had the same flow which needed to call multiple endpoints, and this is that same flow. But I slopped it out when I added it, so it needs some light modificaitons to accept all the fields needed for machine creation (right now I think it only accepts name), and it needs to add PAT auth (right now it only accepts browser auth)

frontend/packages/cli/src/manifest.ts
```diff
@@ -336,13 +337,19 @@ export const commandGroups: CommandGroup[] = [
         format: formatRecord(),
         path: schemas.zGetMachinePath,
       }),
-      op({
+      flowOp({
```
ksarangmath: Once machine setup is a single transactional SDK call (my previous comment), this can go back to using op(). That would preserve --body and --json.

frontend/packages/cli/src/machine-setup.ts
```diff
@@ -0,0 +1,115 @@
...
+  } catch (error) {
+    report.fail(`could not create a daemon token for machine ${machine.id}`)
+    throw error
+  }
+  report.stop('Machine daemon token created')
+  return { machine, token: created.token }
+}
+
+function installCommand(baseUrl: string): string {
+  return `curl -fsSL ${baseUrl}/install/omnarad.sh | sh`
```
ksarangmath: nit: should we validate/shell quote base url

frontend/packages/cli/src/machine-setup.ts
```diff
@@ -0,0 +1,115 @@
...
+  report.url(
+    'Run this on the machine, then paste the machine token when prompted',
+    installCommand(baseUrl),
+  )
+  report.done()
+}
+
+export async function runMachineCreateLocal(context: MachineSetupContext): Promise<void> {
+  const { report, baseUrl } = context
+  const { machine, token } = await createMachineWithDaemonToken(context)
```
ksarangmath: before creating anything, can we detect whether the computer already has an omnara machine configured? and if so, can we just print out to use omnarad start/restart? because if an omnara machine is already configured, then the installer might fail with mismatched token/machine/org etc. ideally the omnara cli create-local only creates an omnara machine for the local computer if nothing exists already, and then points to omnarad

frontend/packages/cli/src/machine-setup.ts
```diff
@@ -0,0 +1,115 @@
...
+  report.info(`Machine ID: ${machine.id}`)
+  report.start('Installing omnarad on this machine')
+  try {
+    await runInstaller(baseUrl, token)
```
ksarangmath: If the current computer doesn't have launchd/systemd, then the installer will launch with the foreground indefinite process. Can we print comments explaining those modes? Foreground mode can say something like "started omnarad in the foreground" and then stream the logs to stdout. The systemd/launchd model mode can print something like "installed and started" cuz it exits after that. and maybe both say use omnarad to manage the omnara daemon, maybe link to docs, etc.

frontend/packages/cli/src/machine-setup.ts
```diff
@@ -0,0 +1,115 @@
...
+  )
+  report.done()
+}
+
+function runInstaller(baseUrl: string, token: string): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const script = [
+      'installer=$(mktemp)',
+      'trap \'rm -f "$installer"\' EXIT',
+      'curl -fsSL "$OMNARA_API_URL/install/omnarad.sh" -o "$installer"',
```
ksarangmath: nit: could we copy the flags from managed_daemon_launcher.sh lines 7-8 for the curl command

frontend/packages/cli/src/manifest.ts
cmsparks: Ah yeah this is a good point, updating this

frontend/packages/cli/src/machine-setup.ts
cmsparks: Yeah this makes a lot more sense, I'll make these backend changes

frontend/packages/cli/src/machine-setup.ts
cmsparks: i'll indicate and provide the PID in this case

api/openapi/openapi.yaml
```diff
@@ -10621,17 +10622,34 @@ components:
     ConnectBYOMachineRequest:
       type: object
       additionalProperties: false
-      required: [display_name, project_ids]
+      required: [display_name]
       properties:
         display_name:
```
ksarangmath, line 10631: Main just merged in a ResourceName type, we can use that here for display name validation

frontend/packages/cli/src/machine-setup.ts
```diff
@@ -0,0 +1,227 @@
...
+  return `curl -fsSL ${shellQuote(`${baseUrl}/install/omnarad.sh`)} | sh`
+}
+
+export const formatMachineSetup: OutputFormat<ConnectByoMachineResponse> = (
+  response,
+  { baseUrl },
+) => ({
+  value: {
+    ...response,
+    install_command: `${installCommand(baseUrl)}  (paste the token when prompted)`,
```
ksarangmath: nit: could we make the install command on its own line so people can easily copy and paste it (i.e. move the (paste the token when prompted) part)

frontend/packages/cli/src/machine-setup.ts
```diff
@@ -0,0 +1,227 @@
...
+  }
+  report.done()
+}
+
+function daemonGuidance(machineId: string): string {
+  return [
+    'omnarad is connected to Omnara as this machine.',
+    'Manage the daemon with: omnarad status | stop | restart | uninstall',
+    `Let agents in a project use this machine with: omnara grant machines add --machine-id ${machineId}`,
+    'Learn more: https://docs.omnara.com/machines/connect',
```
ksarangmath, line 223: nit: docs might need to be updated with any new stuff from this PR

frontend/packages/cli/src/machine-setup.ts
```diff
@@ -0,0 +1,227 @@
...
+    let transcript = ''
+    let foreground = false
+    const consume = (stream: NodeJS.WriteStream) => (chunk: Buffer) => {
+      if (foreground) {
+        stream.write(chunk)
+        return
+      }
+      buffered.push([stream, chunk])
+      transcript += chunk.toString()
+      if (!transcript.includes(noServiceMarker)) return
```
ksarangmath: nit: this seems pretty brittle, could there be a more typed/explicit way of detecting foreground vs systemd/launchd?

frontend/packages/cli/src/machine-setup.ts
```diff
@@ -0,0 +1,227 @@
...
+      settle(() => {
+        reject(new Error(`could not run the omnarad installer: ${error.message}`))
+      })
+    })
+    child.on('close', (code, signal) => {
+      settle(() => {
+        if (foreground && (code === 0 || signal === 'SIGINT')) {
+          report.info('omnarad stopped')
+          resolve()
+        } else if (code === 0) {
```
ksarangmath, line 369: nit: this loses the buffered installer output for the background/systemd/launchd case, which could have some useful info, could we make sure that's printed out?

frontend/packages/cli/src/machine-setup.ts
```diff
@@ -0,0 +1,227 @@
...
+          report.stop('omnarad installed and started as a background service')
+          report.info(daemonGuidance(machineId))
+          resolve()
+        } else {
+          reject(installerFailure(code, signal, foreground ? '' : transcript.trim()))
```
ksarangmath: nit: if the daemon is running in the foreground, and it installs correctly and runs corectly but later exits with an error, it could print a confusing message saying that installation failed and to go through the installation again. We could make it clearer that the daemon failed during runtime, after installation, not an installation failure

frontend/packages/cli/src/machine-setup.ts
```diff
@@ -0,0 +1,227 @@
...
+    report.warn(`omnarad is already installed on this computer: ${existing}`)
+    report.info(
+      [
+        'Nothing was created. Check the daemon with: omnarad status, or start it with: omnarad restart',
+        'To connect this computer as a new machine, run: omnarad uninstall, then re-run this command',
+      ].join('\n'),
+    )
+    throw new CliInputError('omnarad is already installed on this machine')
+  }
+  ensureDaemonHomeWritable(home)
```
ksarangmath, line 190: nit: Before connectMachine(), can we validate that OMNARA_HOME resolves to the same absolute, clean path as omnarad, and that create-local is running on a supported OS/architecture with an API URL the daemon accepts? We can fast fail before creating the remote machine and token.

ksarangmath (approved):
LGTM, just a few nits

frontend/packages/cli/src/machine-setup.ts
cmsparks: yeah ill check the parent PID instead

frontend/packages/cli/src/factory.ts
```diff
@@ -366,7 +369,7 @@ export function registerOperation(parent: Command, config: CliConfig, spec: Oper
         renderResult(data, true)
         return
       }
-      const formatted = spec.format(data as never)
+      const formatted = spec.format(data as never, { baseUrl: config.baseUrl })
```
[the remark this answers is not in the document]
cmsparks: that's correct

PR #409, opened 2026-08-23
fix(auth): canonicalize internationalized email domains

ksarangmath:
## Summary

- canonicalize internationalized email domains to their IDNA ASCII representation at the identity boundary
- recognize equivalent composed Unicode, decomposed Unicode, and Punycode forms when resolving identities and invitations
- serialize verification across equivalent forms and derive normalized values centrally

## Compatibility

- existing composed and decomposed Unicode-domain records remain readable without a data migration
- password login and reset preserve an exact historical identity if equivalent forms already belong to separate accounts
- local-part normalization remains unchanged

## Testing

- `make verify`
- focused storage integration tests for password signup, trusted identity linking, invitations, and concurrent Unicode/Punycode verification`


kvnloo:
CONTRIBUTING asks to open an issue before a substantial change. This PR is +731/−250 across identitystore + generated `identity.sql.go`. I do not see a tracking issue (only open issue is #445 caching).

The change looks like the right identity boundary: IDNA ASCII at write, recognize composed/decomposed/punycode on read, no forced data migration. Tests: `email_normalization_test.go`, `password_auth_store_test.go`, org member integration.

Ask: please confirm `make sqlc-generate` produced `internal/storage/internal/dbsqlc/identity.sql.go` with `internal/storage/queries/identity.sql`. CONTRIBUTING: do not hand-edit generated SQL.

Optional: open a short issue for IDNA identity so this does not look like a drive-by, then I would merge the normalization contract.

PR #411, opened 2026-08-24
feat: Add inline secret adding for mcp

cmsparks:


frontend/apps/web/src/components/agents/AgentConfigMcpSecretDialog.tsx
```diff
@@ -0,0 +1,180 @@
...
+    try {
+      if (loginMode) {
+        await login.start({ name, clientId, clientSecret })
+        return
+      }
+      const material = oauth
+        ? tokenMaterial && { ...tokenMaterial, mcp_url: mcpUrl }
+        : ({ kind: 'generic', value: bearerValue } as const)
+      if (material === undefined) return
+      onCreated(await createSecret.mutateAsync({ owner, name: name.trim(), material }))
```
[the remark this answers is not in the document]
asherdale: i think there might be a separate issue here even when `mcpUrl` is valid: `mcp_url` is only added to encrypted `material`, so `metadata.mcp_url` is empty. since the combobox uses that metadata to match OAuth secrets, should we also pass `metadata: { mcp_url: mcpUrl }`?

frontend/apps/web/src/components/agents/AgentConfigMcpServersField.tsx
```diff
@@ -259,10 +277,102 @@ export function AgentConfigMcpServersField({
...
+  }
+  const actions =
+    server.authType === 'oauth'
+      ? [
+          {
+            label: `Login to ${mcpUrl || 'MCP server'}`,
+            icon: <KeyRound className="size-4 shrink-0" />,
+            disabled: !isMcpOAuthLoginUrl(mcpUrl) || login.pending,
+            onSelect: () => {
+              login.start({ name: defaultMcpSecretName(server) }).catch((error: unknown) => {
```
asherdale: pretty unlikely for this to happen, but just to be safe - should we avoid reusing this generated name when an existing project secret has a different `mcp_url`? same-owner/same-name OAuth login rotates the existing secret in place, so another `github` server at a different URL can silently replace the shared `github-oauth` credential

frontend/apps/web/src/components/agents/AgentConfigMcpServersField.tsx
```diff
@@ -259,10 +277,102 @@ export function AgentConfigMcpServersField({
...
+                setDialog({ error: errorMessage(error, 'Could not start login') })
+              })
+            },
+          },
+          addSecret,
+        ]
+      : server.authType === 'bearer'
```
asherdale: for consistency, should we support creating AWS credentials inline here too? bearer and OAuth now support inline creation, but selecting `SigV4` still requires leaving the builder to create an `aws_credentials` secret first

Image

asherdale (approved):
lgtm

frontend/apps/web/src/components/agents/AgentConfigMcpSecretDialog.tsx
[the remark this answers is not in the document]
cmsparks: mcp_url metadata should be automatically set on the backend, so i think it should be fine

edit: oh i see, in the manual case, that makes sense I'll fix

frontend/apps/web/src/components/agents/AgentConfigMcpServersField.tsx
cmsparks: I'm actually going to just add a little warning indicator if starting auth would overwrite an existing secret, because technically it might be an issue for ones with the same URL as well (I auth with linear MCP on account A, and then it gets overwritten by auth on account B)

PR #412, opened 2026-08-24
fix: surface insufficient Omnara credit errors

asherdale:
- return the categorical “Insufficient Omnara credits.” message for managed-work admission failures across agent launch, managed model requests, and managed machine/process tools
- show organization admins a direct billing link and direct members to an admin across launch and conversation failures, while preserving generic fallbacks when billing is unavailable
- use the existing managed_work_admission_denied code for API and trusted built-in tool failures, and the categorical error text for durable model failures, without reclassifying provider, custom, or MCP errors

### DIFF
--- a/frontend/apps/web/src/components/agents/AgentChatMessage.tsx
+++ b/frontend/apps/web/src/components/agents/AgentChatMessage.tsx
@@ -4,10 +4,15 @@ import { useQuery } from '@tanstack/react-query'
 import { Streamdown } from 'streamdown'
 
 import { AgentAttachment } from '@/components/agents/AgentAttachment'
+import { InsufficientCreditsMessage } from '@/components/agents/InsufficientCreditsMessage'
 import { Brain, Check, ChevronRight, CircleDashed, Terminal } from '@/components/icons'
 import { Bubble, BubbleContent } from '@/components/ui/bubble'
 import { Collapsible, CollapsibleContent, CollapsibleTrigger } from '@/components/ui/collapsible'
 import { Message, MessageContent, MessageHeader } from '@/components/ui/message'
+import {
+  isInsufficientCreditsModelError,
+  isInsufficientCreditsToolError,
+} from '@/lib/insufficient-credits'
 
 function ActorLabel({
   actorID,
@@ -110,16 +115,25 @@ export function AgentChatMessage({
   currentActorId,
   orgID,
   projectID,
+  insufficientCredits,
 }: {
   message: OmnaraUIMessage
   currentActorId?: string
   orgID: string
   projectID: string
+  insufficientCredits?: { billingHref?: string }
 }) {
   const metadata = message.metadata
   const mine =
     message.role === 'user' &&
     (metadata == null || (metadata.actorId != null && metadata.actorId === currentActorId))
+  const showInsufficientCredits =
+    insufficientCredits != null &&
+    message.parts.some(
+      (part) =>
+        (part.type === 'data-model-error' && isInsufficientCreditsModelError(part.data.text)) ||
+        (part.type === 'dynamic-tool' && isInsufficientCreditsToolError(part.toolErrorCode)),
+    )
 
   return (
     <Message align={mine ? 'end' : 'start'}>
@@ -172,6 +186,9 @@ export function AgentChatMessage({
             return <AgentConfigDivider key={part.id} action={part.data.action} />
           }
           if (part.type === 'data-model-error') {
+            if (showInsufficientCredits && isInsufficientCreditsModelError(part.data.text)) {
+              return null
+            }
             return (
               <Bubble key={part.id} align="start" variant="destructive">
                 <BubbleContent className="whitespace-pre-wrap">{part.data.text}</BubbleContent>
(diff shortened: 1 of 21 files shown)


cmsparks:
Don't love the design on the agent profiles page, can we do a persistent toast at the top instead. Something like this:


Screenshot 2026-08-24 at 2 15 43 PM

or this

Screenshot 2026-08-24 at 2 20 04 PM

internal/storage/executionstore/model_outputs_terminal_error.go
```diff
@@ -228,13 +229,19 @@ func publishModelCallErrorOutputTx(
 	if err != nil {
 		return TypedAgentEventRecord{}, err
 	}
+	var metadata resourcemeta.Metadata
+	if contextRow.ErrorKind == modelprotocol.ErrorKindRuntime &&
+		contextRow.ErrorCode == storeerr.ManagedWorkAdmissionDeniedCode {
+		metadata = resourcemeta.Metadata{"omnara_error_code": contextRow.ErrorCode}
```
cmsparks: should this be metadata? IMO metadata is mostly used for display purposes on our end. However, this is error information, which is a fatal error for an agent, blocking it from future work. I think this should probably be it's own strictly typed field. 

This is important especially when external users are consuming the API, because our errors should be strictly typed so a user knows all of the errors they might need to handle

frontend/apps/web/src/components/agents/AgentChatMessage.tsx
```diff
@@ -41,6 +46,23 @@ function JsonValue({ value }: { value: unknown }) {
   )
 }
 
+function isRecord(value: unknown): value is Record<string, unknown> {
```
cmsparks: isRecord() is a code smell, this means that something is probably not sufficiently typed. in particular, isManagedWorkAdmissionDeniedOutput() shouldn't be taking in `unknown` as a type. we already have output.part which is strictly typed, so there's a lot of type checking here that we don't really need to do

frontend/apps/web/src/lib/insufficient-credits.ts
```diff
@@ -0,0 +1,52 @@
...
+  orgID: string,
+  canManage: boolean,
+): Promise<InsufficientCredits | undefined> {
+  try {
+    const data = await queryClient.ensureQueryData(webConfigQuery)
+    return insufficientCredits(data.billingURL, orgID, canManage)
+  } catch {
+    return undefined
+  }
+}
```
cmsparks, line 12: I think these queries are pretty confusing here. It looks like this is just resolving a URL, not actually determining if there are insufficient credits?

Two things I think we should do here:
1) turn this file -> `web-config.ts` and use more accurate naming here.
2) expose one hook called useWebConfig() which resolves billing url and billing href. We can replace the usage from OrganizationNav(). useWebConfig should also return billingHref in the same type.

we should be able to replace this whole file with just this hook
```
type WebConfig {
  billingUrl:
}

export function useWebConfig(): () => Promise<WebConfig> {
  const { activeOrg } = useActiveOrg()
  const { data, isLoading, isError } = useQuery(webConfigQuery)
  return {
    ...data,
    billingHref: (data.billingURL && canManage) ? `${data.billingURL}?org=${orgID}` : undefined,

    // just pipe stuff through from useQuery (I don't actually remember the exact fields, but react query can handle the waits and stuff for us.
    isLoading,
    error
  }
}

File: frontend/apps/web/src/lib/insufficient-credits.test.ts [hunk 4]
```diff
@@ -0,0 +1,46 @@
...
+
+const browser = vi.hoisted(() => ({ fetchWebConfig: vi.fn() }))
+
+vi.mock('@omnara/sdk/browser', () => browser)
+
+beforeEach(() => {
+  browser.fetchWebConfig.mockReset()
+})
+
+describe('resolveInsufficientCredits', () => {
```
cmsparks: I don't think we need this test, react query will handle waiting for inflight queries automatically (from my comment above)

cmsparks (changes requested):
2 main changes, then I can look again: 
* a bit stricter typing both in go + react
* some improvements to the hook semantics here

should end up making the code much simpler to reason about

internal/storage/executionstore/model_outputs_terminal_error.go
asherdale: agreed, i removed the metadata. this keeps the existing typed API code and uses categorical text for durable model errors. as discussed with @ksarangmath, the preference is to not add a strict typing / an error code at this time

frontend/apps/web/src/lib/insufficient-credits.ts
asherdale: good call, added a shared `useWebConfig()` hook and updated all callers

frontend/apps/web/src/components/agents/AgentChatMessage.tsx
```diff
@@ -110,16 +115,25 @@ export function AgentChatMessage({
...
   orgID,
   projectID,
+  insufficientCredits,
 }: {
   message: OmnaraUIMessage
   currentActorId?: string
   orgID: string
   projectID: string
+  insufficientCredits?: { billingHref?: string }
 }) {
```
cmsparks, line 124: nit: we can call useWebConfig() directly inside AgentChatMessage and remove the prop drilling from AgentConversation

cmsparks (approved):
lgtm!

PR #413, opened 2026-08-24
fix: harden Blaxel sandbox pagination

ksarangmath:
This was prompted by repeated failures in the live Blaxel provider smoke test that initially looked flaky. The exact sandbox lookup succeeded, but the bulk observation remained `unknown` until the test timed out.

Investigation showed that Omnara explicitly requested name sorting and Blaxel currently returns overlapping cursor pages for both name sort directions. Repeated copies of the target sandbox were then treated as ambiguous, so the same behavior could also affect production runtime reconciliation.

This removes the unnecessary name sort while retaining the Omnara sandbox filter and cursor pagination. The observer also now collapses identical repeated records defensively. If repeated records disagree, it resolves the ambiguity through Blaxel's exact single-sandbox endpoint.

The focused unit tests, full Go unit suite, broader machine-pool suite, and real Blaxel smoke journey pass. The live journey now completes in about four seconds instead of timing out while waiting for bulk observation.

### DIFF
--- a/internal/machinepool/providers/blaxel/client.go
+++ b/internal/machinepool/providers/blaxel/client.go
@@ -151,7 +151,6 @@ func (c *restClient) ListSandboxes(
 	query := requestURL.Query()
 	query.Set("limit", strconv.Itoa(limit))
 	query.Set("q", "omnara-mch-")
-	query.Set("sort", "name:asc")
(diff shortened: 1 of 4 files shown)


cmsparks (approved):
lgtm

PR #415, opened 2026-08-24
feat: Add MCP registry and server search

cmsparks:
Adds a new service, mcp-registry which does the following:
* at build time, sync from the MCP registry and persist to the built image (this will be rebuilt on every release, so should only get stale occasionally)
* serve new http API which serves content from the MCP registry, which is persisted in a sqlite DB for faster querying

httpapi now:
* serves an HTTP endpoint which essentially proxies to the MCP registry API
* this is a public facing endpoint

frontend now:
* allows searching for mcp servers by name and url to add them in the builder
* serves valid icons for the mcp server
* shows names and descriptions for autofill and in the agent view for servers which are already connected


frontend/apps/web/src/components/mcp/McpServerIdentityGroup.tsx
```diff
@@ -0,0 +1,250 @@
...
+  function select(entry: RegistryServerEntry) {
+    const suggested = registryServerSuggestedName(entry.server)
+    onChange({
+      name: name.trim() === '' && suggested !== '' ? suggested : undefined,
+      url: entry.remote.url,
+    })
+    setAutofilledUrl(entry.remote.url)
+    setOpen(false)
+    setHighlighted(null)
+  }
```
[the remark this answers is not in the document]
cmsparks: This is fine, we don't want to change the name when URL is modified

frontend/apps/web/src/components/mcp/mcpRegistry.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+export function registryServerDomain(server: Pick<McpRegistryServer, 'name'>) {
+  const namespace = server.name.split('/')[0] ?? ''
+  const labels = namespace.split('.').filter((label) => label !== '')
+  if (labels.length < 2 || !labels.every((label) => /^[a-z0-9-]+$/i.test(label))) return null
+  return labels.reverse().join('.').toLowerCase()
+}
+
+export function githubNamespaceUser(server: Pick<McpRegistryServer, 'name'>) {
```
cmsparks, line 17: we do some best effort hacks around fetching images for the server icon, because most servers in the registry don't include an icon

internal/httpapi/spa.go
```diff
@@ -17,7 +17,7 @@ var spaContentSecurityPolicy = strings.Join([]string{
 	"default-src 'self'",
 	"script-src 'self' 'sha256-4Z+0IbR8cDetVQawCZYyJN7DAZJUmjFGeS+nwKwqD8c='",
 	"style-src 'self' 'unsafe-inline'",
-	"img-src 'self' data:",
+	"img-src 'self' data: https:",
```
asherdale, line 20: should we keep icon loading behind an Omnara-controlled boundary instead of allowing every `https:` origin in the CSP and having each browser request publisher icons and guessed favicons directly? this exposes client request metadata to arbitrary third-party origins and can make one search fan out across several of them

the [MCP Registry schema](https://github.com/modelcontextprotocol/registry/blob/main/docs/reference/api/openapi.yaml) also recommends limiting icons to the server domain or another trusted domain and taking precautions with SVGs. i'd rather omit untrusted icons or proxy/cache a vetted subset than relax `img-src` globally

asherdale (changes requested):
the product/UX improvement makes sense, but i have two separate architecture concerns with the current implementation:

1. **the separate image/service:** this adds another artifact to build, deploy, configure, monitor, and keep compatible with the API. the new service also doesn't currently provide an independent refresh lifecycle, so i'm not seeing what the deployment boundary buys us. this functionality could live in the existing API process.

2. **the SQLite store:** even if we want to maintain our own registry snapshot, that doesn't necessarily require a database. the current `streamable-http` catalog at about 778 KB gzipped when reduced to the fields the UI needs. for this size and query pattern, a generated JSON file loaded and searched in memory seems sufficient and would avoid the schema, SQL, and database lifecycle entirely.

there are established examples of both simpler approaches. [VS Code](https://github.com/microsoft/vscode/blob/main/src/vs/platform/mcp/common/mcpGalleryService.ts) queries the MCP registry directly or accepts a JSON file. [Cline](https://github.com/cline/marketplace) publishes a generated `catalog.json`. [Nextcloud](https://github.com/nextcloud/server/blob/master/lib/private/App/AppStore/Fetcher/Fetcher.php) and [OctoPrint](https://github.com/OctoPrint/OctoPrint/blob/master/src/octoprint/plugins/pluginmanager/__init__.py) refresh JSON into their existing process and use file/in-memory caching.

i'd be more comfortable keeping this in the existing API and backing it with either a build-generated filtered JSON snapshot or a periodically refreshed JSON/in-memory cache. that would preserve the UX and our own registry snapshot without requiring either the additional deployment unit or SQLite.

could we simplify this by keeping the registry endpoint in the existing API and starting with a filtered JSON snapshot loaded into memory? if independent refreshes are important, the API could periodically replace that snapshot. this would preserve the same UX while avoiding both a new deployment unit and a database.

internal/httpapi/spa.go
cmsparks: Generally _should_ be safe. There have been CVEs in the past with SVGs, but we don't allow data URLs (which i think is usually the main issue). img-src for foreign origins is a pretty heavily tested part of browsers, so I think it should be ok.

There are some privacy concerns maybe (leaking user agent browser info to some other origin), but I think fetching .ico files is acceptable, I don't think we need to build out a whole proxy just for that.

frontend/apps/web/src/components/mcp/McpServerIdentityGroup.tsx
```diff
@@ -0,0 +1,261 @@
...
+  const [open, setOpen] = useState(false)
+  const [highlighted, setHighlighted] = useState<number | null>(null)
+  const generatedName = useRef('')
+  const lookupToken = useRef(0)
+  const debouncedName = useDebouncedValue(name)
+  const debouncedUrl = useDebouncedValue(url)
+  const filters = registryServerSearchFilters(debouncedName)
+  const searching = filters.q !== undefined
+  const serversQuery = useServers({ filters, pageSize: 25, enabled: open && searching })
+  const results = registryServerEntries(useInfiniteQueryItems(serversQuery))
```
asherdale, line 43: should we filter to active `streamable-http` servers before pagination? for example, `nirholas` has 7 usable matches, but none appear in the first 25 API results, so the UI says “No registry servers match” and hides “Load more”

Dockerfile
```diff
@@ -48,7 +50,9 @@ WORKDIR /app
 
 FROM runtime AS api
 ENV OMNARA_WEB_SERVING=disabled
+ENV OMNARA_MCP_REGISTRY_SNAPSHOT_PATH=/app/mcp-registry/mcp-registry.json
 COPY --from=api-build /out/omnara-api /usr/local/bin/omnara-api
+COPY --from=mcp-registry-snapshot --chown=nonroot:nonroot / /app/mcp-registry/
```
asherdale: should we include the snapshot in source builds too? only `cluster-release.yaml` supplies it, so the documented `docker compose ... --build` flow produces an API whose registry endpoint returns 500

frontend/apps/web/src/components/mcp/mcpRegistry.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+export interface RegistryServerEntry {
+  key: string
+  server: McpRegistryServer
+  remote: McpRegistryRemote
+}
+
+export function streamableRemotes(server: Pick<McpRegistryServer, 'remotes'>) {
+  return server.remotes.filter((remote) => remote.type === 'streamable-http')
```
asherdale, line 77: should we exclude remotes with required headers Omnara can’t currently send? the snapshot includes remotes requiring custom headers such as `X-API-Key`, but the picker offers them even though agent configs only support Bearer, OAuth, and SigV4 auth

asherdale (approved):
lgtm! just a few minor nits

a thought for the future - one thing we'll probably need to add eventually is adding a refresh mechanism here for enterprise/self-hosted deployments that might not deploy as frequently as us

Dockerfile
cmsparks: yeah, missed that. adding

frontend/apps/web/src/components/mcp/McpServerIdentityGroup.tsx
cmsparks: yeah we should filter for just that

frontend/apps/web/src/components/mcp/mcpRegistry.ts
cmsparks: some of the headers included in the results are optional (one instance is cloudflare's MCP server), and it's not required to have any metadata for which headers are optional

PR #416, opened 2026-08-25
fix: prefer hosted Sol for agent templates

asherdale:
- prefer the granted hosted `openai/gpt-5.6-sol` model when initializing or applying agent templates so new template agents do not default to Fable
- preserve the existing oldest-grant fallback when hosted Sol is unavailable or explicitly has tools disabled
- keep manual agent creation available while template defaults load and add focused regression coverage for selection and loading behavior


frontend/apps/web/src/components/agents/CreateAgentForm.test.tsx
```diff
@@ -0,0 +1,230 @@
+/** @vitest-environment happy-dom */
```
ksarangmath: nit: I think this test is more than necessary, can probably be cut down or deleted

ksarangmath (approved):
~~I think we can simplify this PR, since the original reason I wrote the linear ticket changed. But should be pretty straightforward hopefully, or lmk if you think we should still do what this PR does~~

I misunderstood what this PR was doing, the logic is fine, but it can probably be made a bit simpler

frontend/apps/web/src/components/agents/CreateAgentForm.test.tsx
asherdale: removed the test and got rid of the unnecessary robustness

frontend/apps/web/src/components/agents/CreateAgentForm.tsx
```diff
@@ -17,14 +17,19 @@ export function CreateAgentForm() {
...
+  const modelGrants = modelGrantsQuery.data?.pages[0]?.data ?? []
   const defaultPool = (
     poolGrants.find((grant) => grant.machine_pool.management_kind === 'cluster') ?? poolGrants[0]
   )?.machine_pool
-  const defaultModel = modelGrantsQuery.data?.pages[0]?.data[0]?.model
+  const defaultModel =
+    modelGrants.find(
+      ({ model }) =>
+        model.provider_config === 'omnara-openrouter' && model.name === 'openai/gpt-5.6-sol',
+    )?.model ?? modelGrants[0]?.model
```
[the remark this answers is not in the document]
asherdale: this was intentionally removed - reviewer asked to simplify the robustness

PR #419, opened 2026-08-25
fix: Fix API vs APP url issues

cmsparks:
We were using app.omnara.com as the default in the daemon and npm CLI. This fixes that.


frontend/packages/cli/src/config.ts
```diff
@@ -138,7 +157,11 @@ function printConfig(): void {
   console.log(`config file  ${configFilePath()}`)
   console.log(`org_id       ${describeValue('OMNARA_ORG_ID', file.org_id)}`)
   console.log(`project_id   ${describeValue('OMNARA_PROJECT_ID', file.project_id)}`)
-  console.log(`base_url     ${describeValue('OMNARA_BASE_URL', file.base_url, DEFAULT_BASE_URL)}`)
+  console.log(`api_url      ${describeValue('OMNARA_API_URL', file.api_url, DEFAULT_API_URL)}`)
+  console.log(`app_url      ${describeValue('OMNARA_APP_URL', file.app_url, DEFAULT_APP_URL)}`)
+  if (file.base_url !== undefined) {
+    console.log(`base_url     ${file.base_url}  [legacy; replaced by api_url and app_url]`)
+  }
```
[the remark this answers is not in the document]
cmsparks: oh yeah this legacy thing is dumb, i'll fix this

frontend/packages/cli/src/config.ts
```diff
@@ -14,9 +14,12 @@ import {
 } from './interactive.ts'
 import { CliInputError, runCliAction } from './output.ts'
 
-export const DEFAULT_BASE_URL = 'https://app.omnara.com'
+export const DEFAULT_API_URL = 'https://api.omnara.com'
```
ksarangmath: Making this api.omnara.com breaks most CLI commands because the generated SDK has /api/v1 in every path, so it calls https://api.omnara.com/api/v1/..., but Cloudflare exposes those endpoints at https://api.omnara.com/v1/...

I think the cleanest fix is to make api_url the full API root (https://api.omnara.com/v1) and rename app_url to issuer_url. The OpenAPI should be relative (/me, /orgs/..., etc.) instead of hardcoding /api/v1 into every path. Then the same generated client can use https://api.omnara.com/v1 in the CLI and /api/v1 in the frontend.

The Go server can mount the relative handlers centrally under /api/v1 and Cloudflare maps the api.omnara.com /v1 paths to that internal /api/v1 mount.

So the published hosted OpenAPI spec can look something like:
```
servers:
  - url: https://api.omnara.com/v1

paths:
  /me:
  /orgs:
  /orgs/{orgID}/projects:
```

For old CLI configs, we can migrate base_url once. the hosted app.omnara.com value becomes api_url = https://api.omnara.com/v1 and issuer_url = https://app.omnara.com

frontend/packages/sdk/src/device.ts
```diff
@@ -65,6 +65,7 @@ async function postDeviceAuthJSON(
 
 export interface StartDeviceAuthOptions {
   baseUrl: string
+  appUrl?: string
```
ksarangmath: Could this be issuerUrl rather than appUrl? With baseUrl = https://api.omnara.com, the current implementation sends the device request to https://api.omnara.com/api/auth/device/code, but Cloudflare does not route /api/auth from that hostname.

I put up #428, which implements the RFC 8628 device flow and discovers its endpoints from the OAuth issuer. Can you review that PR? I can merge it first, then this PR can be rebased on top and we can release the changes together.

After rebasing, this should pass issuer_url into startDeviceAuth and use the discovered device and token endpoints. The authorization server returns the verification URL as an absolute URL, so the SDK shouldn’t need appUrl or construct /api/auth paths from the API URL.

internal/omnarad/daemon_config.go
```diff
@@ -25,7 +25,7 @@ import (
 const (
 	daemonConfigFileName = "daemon.json"
 	daemonConfigVersion  = 1
-	defaultAPIURL        = "https://app.omnara.com"
+	defaultAPIURL        = "https://api.omnara.com"
```
ksarangmath: this should be changed to https://api.omnara.com/v1 after the above comment changes

frontend/packages/cli/src/machine-setup.ts
```diff
@@ -37,17 +37,17 @@ function shellQuote(value: string): string {
   return `'${value.replaceAll("'", `'\\''`)}'`
 }
 
-function installCommand(baseUrl: string): string {
-  return `curl -fsSL ${shellQuote(`${baseUrl}/install/omnarad.sh`)} | sh`
+function installCommand(apiUrl: string): string {
+  return `curl -fsSL ${shellQuote(`${apiUrl}/install/omnarad.sh`)} | sh`
```
ksarangmath: if the above changes are made, this will need to use the origin of apiUrl, not apiUrl directly, because that'll be suffixed with /v1

frontend/packages/sdk/src/device.ts
ksarangmath: Also issuerUrl is specifically for authentication, so it also shouldn’t be threaded through generic formatting and command-flow contexts as a replacement for appUrl, only this login should need it.

internal/omnarad/daemon_config.go
ksarangmath: Also, since api_url will be the full versioned API root, the daemon client paths should be relative to it (/daemon/... rather than /api/v1/daemon/...). Existing daemon.json files also persist the old https://app.omnara.com value, so we should migrate that exact value to https://api.omnara.com/v1.

ksarangmath (changes requested):
Right now this breaks the CLI because of the way we route things to api.omnara.com, but there's some deeper cleanup we should do that should fix the root cause better

frontend/packages/cli/src/config.ts
```diff
@@ -95,18 +99,48 @@ export function updateConfigFile(patch: Partial<ConfigFile>): ConfigFile {
...
+}
+
+export function resolveUrls(file: ConfigFile, env: NodeJS.ProcessEnv): ResolvedUrls {
+  const issuerUrl = env.OMNARA_ISSUER_URL ?? file.issuer_url ?? DEFAULT_ISSUER_URL
+  const derivedApiUrl =
+    issuerUrl === DEFAULT_ISSUER_URL ? DEFAULT_API_URL : selfHostedApiUrl(issuerUrl)
+  return {
+    apiUrl: env.OMNARA_API_URL ?? file.api_url ?? derivedApiUrl,
+    issuerUrl,
+  }
```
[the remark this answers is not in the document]
cmsparks: this is fine

frontend/packages/cli/src/config.ts
```diff
@@ -138,7 +173,9 @@ function printConfig(): void {
   console.log(`config file  ${configFilePath()}`)
   console.log(`org_id       ${describeValue('OMNARA_ORG_ID', file.org_id)}`)
   console.log(`project_id   ${describeValue('OMNARA_PROJECT_ID', file.project_id)}`)
-  console.log(`base_url     ${describeValue('OMNARA_BASE_URL', file.base_url, DEFAULT_BASE_URL)}`)
+  const { apiUrl, issuerUrl } = resolveUrls(file, {})
+  console.log(`api_url      ${describeValue('OMNARA_API_URL', file.api_url, apiUrl)}`)
+  console.log(`issuer_url   ${describeValue('OMNARA_ISSUER_URL', file.issuer_url, issuerUrl)}`)
```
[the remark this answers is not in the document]
cmsparks: this is probably fine, issuer will eventually be served/defined by oauth metadata, so i think the env var/config value will disappear

frontend/packages/cli/src/config.ts
```diff
@@ -95,18 +99,48 @@ export function updateConfigFile(patch: Partial<ConfigFile>): ConfigFile {
...
+  issuerUrl: string
+}
+
+export function selfHostedApiUrl(issuerUrl: string): string {
+  return `${issuerUrl.replace(/\/+$/, '')}/api/v1`
+}
+
+export function resolveUrls(file: ConfigFile, env: NodeJS.ProcessEnv): ResolvedUrls {
+  const issuerUrl = env.OMNARA_ISSUER_URL ?? file.issuer_url ?? DEFAULT_ISSUER_URL
+  const derivedApiUrl =
```
ksarangmath: nit: can we change this so that issuer_url doesn't implicitly change the API URL, we can just have it be explicit, user supplied issuer url overwrites only issuer url, and user supplied api url only overwrites api url

frontend/packages/sdk/src/client.ts
```diff
@@ -53,7 +53,7 @@ export function createOmnaraClient(options: OmnaraClientOptions = {}): OmnaraCli
     }),
   )
   client.setConfig({
-    baseUrl: options.baseUrl ?? '',
+    ...(options.baseUrl === undefined ? {} : { baseUrl: options.baseUrl }),
```
ksarangmath: nit: when baseUrl is omitted, this client has no base URL even though the OpenAPI spec defines https://api.omnara.com/v1 as the default, could we use that somehow? or just duplicate that default here?

frontend/packages/sdk/README.md
```diff
@@ -14,7 +14,7 @@ npm install @omnara/sdk
 import { createOmnaraClient, bearerToken } from '@omnara/sdk'
 
 const client = createOmnaraClient({
-  baseUrl: 'https://api.omnara.com',
+  baseUrl: 'https://api.omnara.com/v1',
```
ksarangmath, line 17: nit: there are still a few SDK examples using the old origin-only base URLs: examples/cli-agent/src/env.ts:35, examples/x-signal-agent/demo.ipynb:44, and examples/posthog-pulse-agent/demo.ipynb:48. Generated paths no longer include /api/v1, so these might call the wrong routes.

internal/machinepool/providers/managed_daemon_launcher.go
```diff
@@ -48,7 +48,7 @@ func BuildManagedMachineEnv(
 		}
 		env[key] = value
 	}
-	env["OMNARA_API_URL"] = omnaraPublicURL
+	env["OMNARA_API_URL"] = omnaraPublicURL + "/api/v1"
```
ksarangmath: omnaraPublicURL is the app url, so this still gives hosted machines https://app.omnara.com/api/v1 instead of https://api.omnara.com/v1. internal/httpapi/omnarad_install.go:33 makes the same assumption, and managed_daemon_launcher.sh:8 then has to strip a specific suffix to find the installer. Probably should pass the API url and installer urls down directly instead of rebuilding them from publicURL.

ksarangmath (approved):
LGTM, there's some merge conflicts that I think also might add some endpoints that are prefixed with /api/v1 in the openapi spec, so make sure to delete those prefixes when merging if those exist

frontend/packages/cli/src/machine-setup.ts
```diff
@@ -37,17 +37,21 @@ function shellQuote(value: string): string {
   return `'${value.replaceAll("'", `'\\''`)}'`
 }
 
-function installCommand(baseUrl: string): string {
-  return `curl -fsSL ${shellQuote(`${baseUrl}/install/omnarad.sh`)} | sh`
+function installScriptUrl(apiUrl: string): string {
+  return new URL('/install/omnarad.sh', apiUrl).toString()
```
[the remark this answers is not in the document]
cmsparks: this isn't true, api.omnara.com/install/omnarad.sh resolves

frontend/packages/cli/src/config.ts
```diff
@@ -95,18 +99,40 @@ export function updateConfigFile(patch: Partial<ConfigFile>): ConfigFile {
...
+}
+
 export function loadConfig(): CliConfig {
-  const file = readConfigFile()
-  const baseUrl = process.env.OMNARA_BASE_URL ?? file.base_url ?? DEFAULT_BASE_URL
+  const loaded = readConfigFile()
+  const migrated = migrateLegacyBaseUrl(loaded)
+  const file =
+    migrated === undefined ? loaded : updateConfigFile({ ...migrated, base_url: undefined })
+  const { apiUrl, issuerUrl } = resolveUrls(file, process.env)
```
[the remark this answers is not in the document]
cmsparks: this is expected

PR #421, opened 2026-08-26
fix: make agent event streams resilient

ksarangmath:
## The problem

A customer's agent event stream dropped whenever nothing was sent for about 16 seconds. Agent output is bursty, so a healthy turn could go quiet long enough for a proxy to close the connection.

Two more gaps sat behind that one. The TypeScript SDK exposed a single physical connection and left reconnecting, and remembering where to resume, to every caller. On the server, saved events are announced to open streams through Redis; if a notification was missed, the event sat safely in Postgres but stayed undelivered on an otherwise healthy connection.

## What changed

**Server**
- Heartbeat every 10 seconds (was 25), sent as an SSE comment: no event is created and no history is read.
- Each API process runs one small reconciliation loop over its open streams. Redis stays the fast path; Postgres is the source of truth. The loop checks active agents in batches and wakes only streams that are behind, which then catch up through the existing cursor query.

**TypeScript SDK**
- `openAgentEventStream` is one continuous stream across reconnects. It retries transient failures with jittered backoff, honors `Retry-After` (it can lengthen a delay, never shorten it), and resumes from the last delivered saved event with `Last-Event-ID`. Callers write one `for await` loop and get connection-state callbacks for clearing best-effort previews.
- The SDK reads the SSE stream itself instead of through hey-api's SSE helper. That helper runs its own retry loop, hides the response, and swallows errors, so a resilient client had to work around it at every step, and upstream has not changed (hey-api/hey-api#2908, #3018, #4321). URL building, headers, auth, interceptors, types, and zod schemas still come from the generated client.
- A connection with no bytes for 35 seconds, while connecting or afterwards, is treated as stalled and reconnected.

**Clients and docs**
- The React chat, CLI example, quickstart, and notebooks use the SDK stream and clear non-durable previews on reconnect.
- The CLI example prints the turn summary after the answer, keeps tool status visible while tools run, and drops delta-announced tool calls on reconnect.

## Notes
- No API changes; the OpenAPI edits are description text only.
- The SDK composes abort signals with `AbortSignal.any`, so the browser floor is Safari 17.4, Chrome 116, and Firefox 124 (2023–2024). Node 22 already has it.

## Result

Quiet streams stay open through normal idle periods. Temporary disconnects recover on their own. Saved events are eventually delivered even if Redis misses a notification, without turning every heartbeat or connection into a database poller.


frontend/packages/sdk/README.md
```diff
@@ -28,6 +28,38 @@ const client = createOmnaraClient({
 
 Requires Node.js 22 or later.
 
+### Follow an agent event stream
```
cmsparks: we can remove this, already in the docs

frontend/packages/sdk/src/agent-event-stream.ts
```diff
@@ -1,133 +1,386 @@
...
   }
-  return new AgentEventStreamError({
-    kind: 'transport',
-    message: 'Agent event stream disconnected',
-    retryable: true,
-    cause: error,
+  return streamError('transport', 'Agent event stream disconnected', { cause: error })
+}
+
+function isApiErrorBody(data: AgentEventStreamData): data is ApiErrorBody {
```
cmsparks: we shouldn't have these manual validators here, we should be using the zod schemas from the openapi spec. there's some caveats with event streaming, but we should be able to mostly use the openapi zod spec

frontend/packages/sdk/src/agent-event-stream.ts
```diff
@@ -1,133 +1,386 @@
...
+  return (
+    typeof data === 'object' &&
+    'error' in data &&
+    typeof data.error === 'string' &&
+    'code' in data &&
+    typeof data.code === 'string'
+  )
+}
+
+function durableSequence(data: unknown): number | undefined {
```
cmsparks: ditto on the above comment

frontend/packages/sdk/src/agent-event-stream.ts
```diff
@@ -1,133 +1,386 @@
...
+    function finish(completed: boolean) {
+      globalThis.clearTimeout(timeout)
+      signal?.removeEventListener('abort', onAbort)
+      resolve(completed)
     }
+    signal?.addEventListener('abort', onAbort, { once: true })
+  })
+}
+
+function linkedController(signal?: AbortSignal): {
```
cmsparks: we can probably replace this with AbortSignal.any()

```
const controller = new AbortController()
const signal =
  parentSignal == null
    ? controller.signal
    : AbortSignal.any([parentSignal, controller.signal])
```

Technically only has browser support for 2023/4+, but we already expect node 22 support and we can use a polyfill in our frontend.

frontend/packages/sdk/src/agent-event-stream.ts
```diff
@@ -1,133 +1,386 @@
...
+    await response.body.cancel()
+  } catch {
+    // Cleanup is best effort after the owning reader has already been canceled.
   }
-  const error = sourceError()
-  if (error != null) throw error
 }
 
-export async function openAgentEventStream({
+function headersForAttempt(
```
cmsparks: There's a lot of single use 1-5 line functions. Let's try to avoid these (my comment below will probably change a lot of this plumbing anyways though)

frontend/packages/sdk/src/agent-event-stream.ts
```diff
@@ -1,133 +1,386 @@
...
+  headers: StreamEventsData['headers'],
+  acknowledgedSequence: number | undefined,
+): StreamEventsData['headers'] {
+  if (acknowledgedSequence == null) return headers
+  return { ...headers, 'Last-Event-ID': acknowledgedSequence }
+}
+
+export async function* openAgentEventStream({
```
cmsparks, line 261: I like the external facing interface here a lot more, though I don't love the internals of this implementation. Right now it's a lot of imperative functions operating on state that's split up in a few different places.

I think there should be something encapsulating the state here, probably a class, and then that class can implement the Symbol.asyncIterator interface and operate on that state. We can retain the same external interface, openAgentEventStream and it can just yield the iterator. Here's a high level overview of what that might look like:

```ts
export function openAgentEventStream(
    options: OpenAgentEventStreamOptions,
  ): AgentEventStream {
    return new AgentEventStream(options)
  }

  export class AgentEventStream implements AsyncIterable<AgentEventStreamFrame> {
    readonly #options: OpenAgentEventStreamOptions
    readonly #controller = new AbortController()
    #iterator: AsyncGenerator<AgentEventStreamFrame> | undefined
    #acknowledgedSequence: number | undefined
    #connectionState: AgentEventStreamConnectionState | undefined

    constructor(options: OpenAgentEventStreamOptions) {
      this.#options = options
    }

    get lastEventId(): number | undefined {
      return this.#acknowledgedSequence
    }

    get connectionState(): AgentEventStreamConnectionState | undefined {
      return this.#connectionState
    }

    close(): void {
      this.#controller.abort()
    }

    [Symbol.asyncIterator](): AsyncIterator<AgentEventStreamFrame> {
      this.#iterator ??= this.#run()
      return this.#iterator
    }

    async *#run(): AsyncGenerator<AgentEventStreamFrame> {
      const controller = new AbortController()
      const signal =
        parentSignal == null
          ? controller.signal
          : AbortSignal.any([parentSignal, controller.signal])
      let consecutiveFailures = 0
      let hasRetried = false

      try {
        while (!signal.aborted) {
          const attempt = new StreamAttempt({
            client: this.#options.client,
            path: this.#options.path,
            query: this.#options.query,
            headers: headersForAttempt(this.#options.headers, this.#acknowledgedSequence),
            parentSignal: signal,
            hasRetried,
            onConnected: (state) => this.#notify(state),
          })

          try {
           // insert other retry stuff
           // ...
           // ...
           // ...

          this.#setState({
            state: 'reconnecting',
            attempt: consecutiveFailures,
            delayMs,
            error: attempt.failure,
          })
          hasRetried = true
          if (!(await abortableDelay(delayMs, signal))) return
        }
      } finally {
        unlink()
      }
    }

    // could alternatively be a setter function
    #setState(state: AgentEventStreamConnectionState): void {
      this.#connectionState = state
      this.options.onConnectionStateChange(state)
    }
  }
```

Usage ends up being the same, like:

```ts
const stream = openAgentEventStream({ client, path })
for await (const frame of stream) {
  handleFrame(frame)
}
// from anywhere else — a timeout, UI unmount, another task:
stream.close()
```


I think this will dramatically simplify the state handling and callback handlers, and make the code easier to reason about.

cmsparks (commented):
The reconciler design + golang stuff lgtm, though I have some feedback on the agent-event-stream implementation, now that that file is owning more of the client functionality.

frontend/packages/sdk/README.md
ksarangmath: Yup, removed all edits to readme

frontend/packages/sdk/src/agent-event-stream.ts
ksarangmath: using the zod schema for checking the frames now

frontend/packages/sdk/src/agent-event-stream.ts
ksarangmath: Yup good call, done

frontend/packages/sdk/src/agent-event-stream.ts
ksarangmath: Got rid of most of these in the refactor

frontend/packages/sdk/src/agent-event-stream.ts
ksarangmath: Yeah this was messy cuz it was handling the agent event stream state and the physical connection state. I restructured this so that those are clearly separate, with their own state. And the stream.close in your example is done with an abort signal, which is what all the consumers should use. lmk if this looks good or if there's changes I should make

frontend/packages/sdk/src/agent-event-stream.ts
```diff
@@ -1,133 +1,331 @@
...
+
+function notifyConnectionState(
+  callback: OpenAgentEventStreamOptions['onConnectionStateChange'],
+  state: AgentEventStreamConnectionState,
+): void {
+  try {
+    callback?.(state)
+  } catch (error) {
+    throw streamError('client', 'Agent event stream connection callback failed', { cause: error })
   }
```
cmsparks: this feels like excessive error handling, i don't think we need a try catch for the callback handler. the user provides/writes the callback, so they will know if their callback fails. we should just directly call onConnectionStateChange(), can we replace all callsites for notifyConnectionState with instead just directly calling onConnectionStateChange?.(...stateParams)

frontend/packages/sdk/src/agent-event-stream.ts
```diff
@@ -1,133 +1,331 @@
...
+    }
+    if (response.body == null) {
+      throw streamError('contract', 'Agent event stream response has no body')
+    }
+    reader = response.body.getReader()
+    signal.throwIfAborted()
+    yield { kind: 'connected' }
+    const parser = createServerSentEventParser()
+    const decoder = new TextDecoder()
+    for (;;) {
```
cmsparks: nit: while (true) {

(i think this is caused by a eslint rule sometimes)

frontend/packages/sdk/src/validate-response.ts
```diff
@@ -40,8 +40,16 @@ export function isUnknownEnumError(error: z.ZodError, data: unknown): boolean {
   return error.issues.every((issue) => isUnknownEnumIssue(issue, data))
 }
 
-export async function validateResponse(schema: z.ZodType, data: unknown): Promise<void> {
+export async function schemaMismatch(
```
cmsparks: We should remove this function because it doesn't actually enforce schemas in a typesafe way. Instead, we should directly call schema.parse() inline at all the callistes instead of putting this as a helper function.

If you look at `data` at the callsites here, you can see that it's still marked as unknown after calling the function and then we subsequently cast it. Technically the code _might_ be correct, but we don't really get any of the typescript typesafety guarantees. If we call schema.parse/safeparse() we preserve the datatype returned and get clearer typesafety

(fwiw we're pretty good at manually fixing these because we do a lot of generated code, but I need to add lint rules for this so we prevent these things from cropping up in our codebase)

frontend/packages/sdk/src/validate-response.ts
```diff
@@ -40,8 +40,16 @@ export function isUnknownEnumError(error: z.ZodError, data: unknown): boolean {
...
+  data: unknown,
+): Promise<z.ZodError | undefined> {
   const result = await schema.safeParseAsync(data)
-  if (result.success || isUnknownEnumError(result.error, data)) return
-  throw result.error
+  if (result.success || isUnknownEnumError(result.error, data)) return undefined
+  return result.error
+}
+
+export async function validateResponse(schema: z.ZodType, data: unknown): Promise<void> {
```
cmsparks: also remove this function for the same reason as above

frontend/packages/sdk/src/validate-response.ts
cmsparks: I'm gonna add this in a PR later today: https://github.com/dmmulroy/anti-slop

which should catch these things

cmsparks (approved):
lgtm, mostly just nits with typesafety

frontend/packages/sdk/src/agent-event-stream.ts
cmsparks: I think this rework looks fine, i'm fine with the separation of open connection and retry loop even if i slightly prefer the class stylistically

frontend/packages/sdk/src/agent-event-stream.ts
ksarangmath: Yeah good catch, this doesn't really do anything. removing it

frontend/packages/sdk/src/validate-response.ts
ksarangmath: I ended up making a helper that returns the the same shape as safeparse (uses safeparse internally), and it also checks the enum forward compat. The alternative was to make enums more broad in the codegen part, was split 50/50 but ended up going with this cuz I felt it was more explicit, and the broadness is only needed for responses, not requests.

PR #422, opened 2026-08-26
build: adopt a bounded Dependabot maintenance policy

ksarangmath:
## Summary

- run Dependabot on a predictable Monday morning schedule with at most three open version-update PRs per ecosystem
- group routine GitHub Actions, related Go, and npm production/development minor/patch updates
- group security updates separately from the routine version-update queue
- cover both pnpm workspaces and keep TypeScript updates aligned across them
- add Docker coverage and group the same image across directories
- keep Node and Go toolchain line changes manual while allowing digest and patch refreshes

## Policy

Routine minor/patch work is grouped to reduce queue volume. Higher-risk major updates remain separate unless the repository has an explicit compatibility line: Node and Go container line changes are intentionally ignored until a human updates that line.

The frontend remains on pnpm 11.22. Dependabot core currently lists pnpm 11 as supported and includes pnpm-11-specific update logic, even though GitHub's published ecosystem table has not yet caught up.

This configuration also defines security-update grouping. Automatic Dependabot security-update PRs are a separate repository setting; GitHub's status endpoint confirms they are currently disabled, so enable them after this configuration lands.

## Validation

- `git diff --check`
- `uvx check-jsonschema --schemafile https://json.schemastore.org/dependabot-2.0.json .github/dependabot.yml`
- `make web-check`

## References

- [GitHub: optimizing Dependabot PR creation](https://docs.github.com/en/code-security/tutorials/secure-your-dependencies/optimizing-pr-creation-version-updates)
- [GitHub: Dependabot options reference](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference)
- [GitHub: Dependabot comment commands](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-pull-request-comment-commands)
- [Dependabot's own configuration](https://github.com/dependabot/dependabot-core/blob/main/.github/dependabot.yml)
- [Dependabot core pnpm support](https://github.com/dependabot/dependabot-core/blob/main/npm_and_yarn/lib/dependabot/npm_and_yarn/pnpm_package_manager.rb)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,8 +4,21 @@ updates:
     directory: /
     schedule:
       interval: weekly
+      day: monday
+      time: "09:00"
+      timezone: America/Los_Angeles
+    open-pull-requests-limit: 3
+    groups:
+      actions-minor-and-patch:
+        applies-to: version-updates
+        patterns:
+          - "*"
+        update-types:
+          - minor
+          - patch
     cooldown:
       default-days: 7
+
   - package-ecosystem: gomod
     directories:
       - /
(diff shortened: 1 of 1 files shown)


kvnloo:
CONTRIBUTING check: one problem, conventional title `build:`. Generated files untouched.

Nit on `.github/dependabot.yml`: npm security groups `production-security-updates`, `development-security-updates`, and `other-npm-security-updates` with `patterns: ["*"]`. Dependabot assigns a dependency to the first matching group, so the catch-all is last-wins only if GitHub evaluates in list order — worth confirming, because a catch-all `*` security group can hide whether prod vs dev actually split.

`open-pull-requests-limit: 3` per ecosystem plus several version-update groups means a Monday batch can fill with grouped minor/patch PRs while security waits. Docker ignores golang major+minor but node only major; that matches the PR body.

I would merge if the intent is bounded Dependabot. Optional non-blocker: drop the npm catch-all security group or give security groups a higher priority / separate limit so they cannot queue behind grouped typescript bumps.

PR #425, opened 2026-08-26
feat: add artifact upload and display tools

asherdale:
- add built-in `upload_artifact`, backed by the existing managed-process lifecycle, so agents can upload files from attached machines without a parallel daemon dispatch system
- add a daemon-authenticated upload endpoint that validates the active tool call and machine, accepts non-empty regular files up to 10 MiB, and stores agent-scoped artifacts
- add built-in `show_artifact` structured results plus inline dashboard cards for safe image previews and artifact downloads


internal/harness/tools/artifact_tools.go
```diff
@@ -0,0 +1,211 @@
...
+		toolCallID,
+		encodedPath,
+	)
+	return executionstore.CreateProcessInput{
+		IOMode:         processcmd.IOModePipe,
+		Command:        command,
+		ShellSelector:  processcmd.ShellDefault,
+		InitialWaitMS:  processaction.MaxWaitMilliseconds,
+		TimeoutSeconds: uploadArtifactProcessTimeoutSeconds,
+	}
```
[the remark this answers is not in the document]
asherdale: not relevant currently - `omnarad` does not run on windows yet

internal/harness/tools/artifact_tools.go
```diff
@@ -0,0 +1,211 @@
...
+	toolCallID, err := publicid.Encode(publicid.KindToolCall, call.ToolCallID)
+	if err != nil {
+		return nil, fmt.Errorf("encode tool call id: %w", err)
+	}
+	authorizationInput, err := uploadArtifactAuthorizationInput(binding.ID, resolved.Path)
+	if err != nil {
+		return nil, err
+	}
+	return startProcessTool(
```
asherdale, line 99: some architectural context: `upload_artifact` intentionally runs through the existing process lifecycle instead of introducing a separate daemon request/reply flow.

this lets it reuse the established machine-selection, wakeup, timeout, cancellation, persistence, observability, and idle-activity behavior. it also means uploads produce normal process output and failures that can be inspected with `read_process`, rather than creating another asynchronous lifecycle alongside processes.

the hidden `omnarad` subcommand is the process payload responsible for reading the sandbox file and making the authenticated upload. keeping that operation inside the trusted daemon helper avoids exposing machine credentials to the agent command while also avoiding a new daemon message type, reply channel, outbox, or artifact-specific lifecycle.

docs/events/artifacts.mdx
```diff
@@ -4,14 +4,39 @@ description: View metadata and download files referenced in an agent's event log
...
 
 ```json
 { "type": "media_ref", "artifact_id": "art_z3jehcyd5n6a2bfgik7mv4qtrw" }
 ```
 
 In the dashboard, artifacts appear as downloadable attachment chips on the conversation timeline; rendering your own timeline means resolving each `media_ref` with the two endpoints on this page.
 
+## Show an existing artifact
+
+A successful [`show_artifact`](/tools/built-in#show_artifact) result carries artifact metadata in a `structured_data` block rather than a `media_ref`, so presenting it does not make its bytes eligible for later model context. Match the result's `tool_call_id` to the corresponding `show_artifact` call, then use its `artifact_id` with the endpoints below.
```
cmsparks: this doesn't really show how to use this tool, i think we could include an explanation like:

```
The show artifact tool is a tool intended to be used to show a file to an end user. For example, if the agent generated a PDF and would like to display it for an end user, it is impossible or inefficient for the agent to reproduce the file in it's context. The agent will explicitly call the show_artifact tool referencing a file.

In your client, you should resolve the artifact using the artifact id referenced in the tool result. 
```

cmsparks: > result carries artifact metadata in a `structured_data` block rather than a `media_ref`, so presenting it does not make its bytes eligible for later model context

I think that we should have a more specific specific type for the tool result here, rather than putting it in `structured_data`.

I think there's two options:
1. extend media_ref to add a new parameter (i.e. `model_context: 'include' | 'exclude'`) to distinguish whether the content should be included in context. IMO I lean towards this because it's a bit more expressive as a type. Then when users want to display artifact content, they only need to match on one type as opposed to two separate types.
2. add a new content_block type for this specific variant

This will make it so that users don't need to create helper functions like shownArtifactFromToolPart(part: DynamicToolPart) which does a lot of ugly typechecks and instead can just directly use the tools

internal/httpapi/middleware.go
```diff
@@ -33,6 +33,8 @@ const maxAttachmentRequestBodyBytes int64 = 2 * modelcontext.MaxResolvedMediaByt
 
 const maxSkillUploadRequestBodyBytes int64 = int64(skills.MaxArchiveBytes) + 1024*1024
 
+const maxDaemonArtifactUploadBytes int64 = 10 * 1024 * 1024
```
cmsparks: nit: can we define the maxDaemonArtifactUploadBytes in one place, we also define it in artifact_upload.go in the daemon as maxArtifactUploadBytes

docs/tools/built-in.mdx
```diff
@@ -159,6 +160,16 @@ Deletes a pool-backed machine. Requires `machine_ref`.
 
 `list_machines` lists the agent's BYO and pool-backed machines. `inspect_machine` shows whether a specific machine is ready to run commands; provide `machine_ref` when the agent has more than one. Results identify the machine's `source_kind` and `binding_kind`; only pool bindings can be passed to `delete_machine`.
 
+## Artifacts
+
+### `upload_artifact`
```
cmsparks, line 165: is there ever a world where we're uploading an artifact, but not showing it? i don't think so right, because there's not really anything else that the agent can do with it?

cmsparks (approved):

docs/events/artifacts.mdx
asherdale: good call, added `exclude_from_model_context` to `content_blocks` with default of `false`

docs/tools/built-in.mdx
asherdale: yea you're right, i collapsed both tools into `upload_artifact`

internal/storage/executionstore/tool_calls_workflow.go
```diff
@@ -233,6 +234,58 @@ func commandTerminalToolResult(
...
+	}
+	if err != nil {
+		return "", nil, fmt.Errorf("load uploaded artifact: %w", err)
+	}
+	contentParts, err := marshalJSON([]map[string]any{{
+		"type":                       "media_ref",
+		"artifact_id":                artifact.ID.String(),
+		"exclude_from_model_context": true,
+	}})
+	return ToolResultOutcomeSucceeded, contentParts, err
```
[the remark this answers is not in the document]
asherdale: not true - the model-context projection filters these blocks out, confirmed in a live spike

PR #427, opened 2026-08-26
fix: reject database-unsafe content before storage

asherdale:
- reject database-unsafe U+0000 and invalid UTF-8 before PostgreSQL or blob writes, returning non-retryable validation errors for malformed public requests
- canonicalize structured-data JSON before storage so unsupported surrogate escapes become replacement characters while nested U+0000 keys and values are rejected
- preserve retry and idempotency behavior: rejected tool results remain ready for correction, artifact uploads avoid partial writes, and completed launch replays keep their existing semantics
- document the custom tool-result content contract for client implementers

### DIFF
--- a/docs/tools/custom.mdx
+++ b/docs/tools/custom.mdx
@@ -119,6 +119,8 @@ curl "$OMNARA_API/orgs/$ORG/projects/$PROJ/agents/$AGENT/tool-calls/tcl_wz3jehcy
 
 The agent wakes and continues its turn with your result in context. If your system failed, submit `outcome: "failed"` with a text block explaining why — a failed tool call is information the model can act on, not an error you should swallow.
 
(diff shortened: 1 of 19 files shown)


docs/tools/custom.mdx
```diff
@@ -119,6 +119,8 @@ curl "$OMNARA_API/orgs/$ORG/projects/$PROJ/agents/$AGENT/tool-calls/tcl_wz3jehcy
 
 The agent wakes and continues its turn with your result in context. If your system failed, submit `outcome: "failed"` with a text block explaining why — a failed tool call is information the model can act on, not an error you should swallow.
 
+Result text, structured-data string keys and values, and metadata must not contain U+0000. Decode text output before submitting it, and send binary output as a supported media block or a safe failed result. An invalid result returns `400` and leaves the call `ready`; correct the payload instead of retrying the same payload.
```
cmsparks: I think we can remove everything after the first sentence. "Result text, structured-data string keys and values, and metadata must not contain U+0000." should be sufficient

cmsparks (approved):
lgtm

PR #428, opened 2026-08-26
feat: implement OAuth 2.0 device authorization grant

ksarangmath:
## Summary

- implement the RFC 8628 device authorization grant for CLI login
- publish RFC 8414 authorization-server metadata and discover endpoints in the SDK
- bind device codes to the registered public CLI client and issue normal Omnara personal access tokens
- preserve device approval through login or signup before organization onboarding
- move generated API clients to the canonical public API base while keeping OAuth and browser flows on the app issuer
- use the public API origin for daemon installation and managed-machine bootstrap

## Migration and routing

- this is an intentional coordinated CLI hard cutover; current users are directed to use `npx omnara@latest`
- hosted clients use `https://api.omnara.com/v1` as the API base and `https://app.omnara.com` as the OAuth issuer
- the installer is served at `https://api.omnara.com/install/omnarad.sh` through the existing Cloudflare API-origin rule
- `OMNARA_PUBLIC_API_URL` configures the complete public API base for the API, worker, and maintenance services; self-hosted deployments default to `OMNARA_PUBLIC_URL + /api/v1`
- CLI config now stores `api_url` and `issuer_url`; the legacy `base_url` key is removed
- persisted legacy daemon API origins are migrated in memory, while explicit `OMNARA_API_URL` values are treated as complete API bases


internal/httpapi/auth/handler.go
```diff
@@ -169,6 +170,7 @@ func httpClientWithoutRedirects(client *http.Client) *http.Client {
 }
 
 func (h *Handler) RegisterRoutes(mux *http.ServeMux) {
+	mux.HandleFunc("GET /.well-known/oauth-authorization-server", h.authorizationServerMetadataRoute)
```
cmsparks, line 173: This will be useful assuming we host an MCP server for our API

internal/httpapi/auth/oauth_server.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func (h *Handler) authorizationServerMetadataRoute(w http.ResponseWriter, r *http.Request) {
+	issuer := h.issuerURL(r)
+	if issuer == "" {
+		writeOAuthError(w, http.StatusServiceUnavailable, "temporarily_unavailable", "issuer is not configured")
+		return
+	}
+	w.Header().Set("Cache-Control", "public, max-age=3600")
+	writeJSON(w, http.StatusOK, map[string]any{
```
cmsparks, line 42: https://datatracker.ietf.org/doc/html/rfc8414

RFC8414 marks `response_types_supported` as "REQUIRED"

migrations/000027_device_oauth_client.sql
```diff
@@ -0,0 +1,8 @@
+-- +goose Up
+
+ALTER TABLE auth_device_flows
+    ADD COLUMN client_id text NOT NULL DEFAULT 'omnara-cli',
+    ADD CONSTRAINT auth_device_flows_client_id_check CHECK (
+        char_length(client_id) BETWEEN 1 AND 256
+        AND client_id !~ '[[:cntrl:]]'
```
cmsparks: tiny nit: This isn't technicallly the same check as the ` strings.ContainsFunc(` check. should match only control characters in the go code or do a stricter check here

cmsparks (approved):
lgtm

internal/httpapi/auth/oauth_server.go
ksarangmath: https://errata.rfc-editor.org/eid7793/ - There's an correction for this case, since we don't have an authorization endpoint and just have device grant, we should leave this as is I think:

'At the same time, section 3.2 says that "Claims with zero elements MUST be omitted from the response."'

ksarangmath: But we probably need to add this when we do MCP OAuth

migrations/000027_device_oauth_client.sql
ksarangmath: Yeah good catch, will fix this

migrations/000028_device_oauth_client.sql
```diff
@@ -0,0 +1,11 @@
...
+
+ALTER TABLE auth_device_flows
+    ADD COLUMN client_id text NOT NULL DEFAULT 'omnara-cli',
+    ADD CONSTRAINT auth_device_flows_client_id_check CHECK (
+        octet_length(client_id) BETWEEN 1 AND 256
+        AND (client_id COLLATE "C") !~ '[[:cntrl:]]'
+    );
+
+ALTER TABLE auth_device_flows
+    ALTER COLUMN client_id DROP DEFAULT;
```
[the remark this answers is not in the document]
ksarangmath: Fixed in 7ddccc84: the migration now retains the omnara-cli default so the previous API writer remains compatible, with an integration test covering the old insert shape. We can drop the default in a later release once that server is no longer a rollback target.

frontend/packages/cli/src/login.ts
```diff
@@ -159,8 +163,12 @@ export function registerLoginCommand(program: Command, cli: CliConfig): void {
           throw error
         }
         report.stopWaiting(true)
-        const saved = updateConfigFile({ token, base_url: cli.baseUrl })
-        const client = createOmnaraClient({ baseUrl: cli.baseUrl, auth: bearerToken(token) })
+        const saved = updateConfigFile({
+          token,
+          api_url: cli.apiUrl,
+          issuer_url: cli.issuerUrl,
+        })
```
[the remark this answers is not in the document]
ksarangmath: Fixed in 3dc11e7. Login now performs the real config write before starting device authorization, so an invalid or unwritable config fails before a PAT can be minted.

frontend/apps/web/src/components/overview/onboardingCli.ts
```diff
@@ -64,7 +64,6 @@ export function chatCommands(input: {
   const sdk = `import { bearerToken, createOmnaraClient, sdk } from '@omnara/sdk'
 
 const client = createOmnaraClient({
-  baseUrl: '${input.origin}',
```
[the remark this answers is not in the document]
ksarangmath: Fixed in 3dc11e7. The existing web-config endpoint now exposes the deployment’s canonical public API URL, and both the SDK and curl examples use that same value.

PR #430, opened 2026-08-27
fix: guide agents before creating machines

asherdale:
- guide agents to call `list_machines` before requesting pool capacity, reusing executable machines and waiting for suitable machines that are still provisioning
- retain the existing `machine_pool_name` guidance for multi-pool configurations

### DIFF
--- a/internal/toolcatalog/catalog.go
+++ b/internal/toolcatalog/catalog.go
@@ -32,7 +32,7 @@ const (
 	stopProcessToolDescription    = "Interrupt or terminate a running process."
 	readProcessToolDescription    = "Read retained process output, optionally waiting briefly for output or completion."
 	listProcessesToolDescription  = "List active processes in the current agent, including process_id values."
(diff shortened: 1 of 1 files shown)


internal/toolcatalog/catalog.go
```diff
@@ -32,7 +32,7 @@ const (
 	stopProcessToolDescription    = "Interrupt or terminate a running process."
 	readProcessToolDescription    = "Read retained process output, optionally waiting briefly for output or completion."
 	listProcessesToolDescription  = "List active processes in the current agent, including process_id values."
-	createMachineToolDescription  = "Request another pool-backed machine for this agent. machine_pool_name is only needed when multiple machine pools are available."
+	createMachineToolDescription  = "Request a pool-backed machine for this agent. First call list_machines to check for a suitable existing machine; use it if executable or wait for it if provisioning. machine_pool_name is only needed when multiple machine pools are available."
```
ksarangmath: nit: should we say something like "call list_machines if you have access to the tool"? In the off chance that list machines isn't added, but that'd be pretty rare

ksarangmath (approved):
LGTM

PR #440, opened 2026-08-28
feat: add download_artifact tool

asherdale:
- add a built-in `download_artifact` tool that copies conversation artifacts onto an attached machine through the existing process lifecycle
- authorize daemon downloads against the active tool call, selected machine, agent scope, and exact public artifact ID
- expose public artifact IDs to models while keeping artifact bytes excluded from later model context
- download through a hidden `omnarad` helper with relative and `~` paths, atomic replacement, bounded transfers, and preserved destination permissions
- wire the tool into defaults, permission prompts, API warnings, generated clients, and artifact documentation


internal/model/anthropicmessages/request.go
```diff
@@ -316,6 +316,9 @@ func renderAnthropicContent(
 			artifactID := jsonFieldText(part["artifact_id"])
 			if resolved, ok := media[artifactID]; ok {
 				if block, ok := anthropicMediaBlock(resolved); ok {
+					if ref := modelcontext.ArtifactPublicID(artifactID); ref != "" {
```
cmsparks, line 319: Should we abbreviate artifact IDs like we do for machines?

internal/omnarad/artifact_download.go
```diff
@@ -0,0 +1,118 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/omnara-ai/omnara/internal/daemonprotocol"
+	"github.com/omnara-ai/omnara/internal/processcmd"
+	"github.com/omnara-ai/omnara/internal/publicid"
+)
+
+func runDownloadArtifactCommand(
```
cmsparks, line 20: not related to your PR, but in hindsight, this is a much easier way to implement downloading, we should probably swap the skills API to download through a hidden daemon command, because technically it can fail currently.

cmsparks (approved):
lgtm

internal/omnarad/artifact_download.go
asherdale: good call, created a linear ticket for this

internal/model/anthropicmessages/request.go
asherdale: definitely worth considering - especially if models struggle with replicating the IDs exactly. but this would be a much larger change (need a migration, uniqueness handling, extra queries, etc) touching a lot of pre-existing code, so leaving as a follow-up PR for now

PR #442, opened 2026-08-28
fix: reduce production health noise and maintenance load

asherdale:
- Classify client-disconnected API requests as empty 499 responses with cancellation telemetry instead of reporting them as server failures.
- Keep idle worker polls at debug level while preserving info and error visibility for completed work, recoverable races, and genuine failures.
- Replace the all-project wakeup rebuild N+1 loop with bounded global agent batches that skip deleted projects and retain the existing project-scoped path.
- Suppress only operation errors that are confirmed shutdown cancellations across core and machine-pool maintenance while preserving genuine runtime-lock and reconciliation failures.


internal/httpapi/openapi_strict.go
```diff
@@ -59,10 +61,30 @@ func writeOpenAPIRequestError(w http.ResponseWriter, r *http.Request, err error)
 }
 
 func openAPIResponseErrorHandler(w http.ResponseWriter, r *http.Request, err error) {
+	if handleCanceledRequest(w, r) {
```
ksarangmath: Could we also check whether err was caused by the cancellation? A client disconnect explains the 499, but an unrelated handler or storage failure could happen at the same time. In that case, we should still record 499 while preserving the real error instead of downgrading it to info.

ksarangmath: I think request-outcome classification belongs at the outer http.request boundary rather than individual OpenAPI/auth error exits. Client disconnection is a transport outcome, while the handler/auth result is an operation outcome; they are independent and can occur together. Today handleCanceledRequest ignores err, and ResponseError drops its original cause, so a genuine failure racing a disconnect can become info/499 and disappear.

Could we preserve the operation cause through API error mapping, have lower layers report their operation outcome without deciding the final request severity, and resolve the two outcomes once when http.request is finalized? A cancellation-only operation would become client_closed/info, while any genuine application result would keep its normal status and severity with the disconnect recorded separately. That would also make 499 derived transport telemetry rather than an application response contract and avoid repeating cancellation policy across every error path.

internal/log/logent/maintenance.go
```diff
@@ -33,12 +34,20 @@ func MaintenanceLoopResult(
...
-		log.Error(ctx, rebuildErr)
+		if !IsCanceledDuringShutdown(ctx, rebuildErr) {
+			fields["maintenance.rebuild_agent_wakeups.error"] = rebuildErr.Error()
+			log.Error(ctx, rebuildErr)
+		}
 	}
 	log.Attach(ctx, fields)
 }
+
+func IsCanceledDuringShutdown(ctx context.Context, err error) bool {
```
ksarangmath: IsCanceledDuringShutdown assumes that an error matching context.Canceled represents a cancellation-only outcome, but the producers do not guarantee that. In particular, runBoundedReconcile keeps only the first concurrent error, so a canceled operation can win while a later genuine provider or storage failure is discarded; this helper then suppresses the result completely.

Could the reconciliation boundary preserve interruption separately from genuine failures, retaining the relevant failures from concurrent work, and let the maintenance owner classify the completed outcome once? Then cancellation-only shutdown can remain quiet without relying on logging code to infer whether an aggregate error contained real failures.

internal/storage/executionstore/agent_wakeups_store.go
```diff
@@ -59,19 +55,5 @@ func (s *Store) RebuildMissingAgentWakeups(
...
-	for _, projectID := range projectIDs {
-		count, err := s.RebuildMissingAgentWakeups(ctx, projectID)
-		if err != nil {
-			rebuildErrs = append(rebuildErrs, fmt.Errorf("project %s: %w", projectID, err))
-			continue
-		}
-		total += count
-	}
-	return total, errors.Join(rebuildErrs...)
+	return s.rebuildMissingAgentWakeups(ctx, nil)
```
ksarangmath: Rebuilding the agent wakeups is something that was added in case a wakeup happens to now get created when an input comes in, but if we guarantee that every input creates a wakeup, we could remove the rebuilds altogether. Maybe that's the cleaner solution, rather than running this maintenance loop?

ksarangmath: If we do keep the loop, this still scans every agent before returning, then repeats the full scan on the next maintenance run. Could it process a fixed amount of work each time and continue where it left off on the next run? We should also make sure one failing batch can’t cause every later agent to be skipped indefinitely.

internal/storage/executionstore/agent_runtime_lock_reaping_store.go
```diff
@@ -39,6 +39,9 @@ func (s *Store) ReapExpiredAgentRuntimeLocks(ctx context.Context, batchSize int3
 			candidate.ID,
 		)
 		if err != nil {
+			if ctxErr := ctx.Err(); ctxErr != nil && errors.Is(err, ctxErr) {
+				break
+			}
```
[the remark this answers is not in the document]
asherdale: valid, fixed with `errutil.OnlyMatches` so mixed failures are preserved

internal/httpapi/openapi_strict.go
asherdale: good call, fixed this by moving cancellation classification to the `http.request` finalizer. lower layers now report their mapped operation outcome, `ResponseError` preserves its cause, and `499` is telemetry-only. only cancellation-only errors are demoted, and genuine failures or panics racing a disconnect retain their normal status and severity

internal/log/logent/maintenance.go
asherdale: fixed with the same approach: preserve mixed/genuine failures, suppress only cancellation-only outcomes, and classify once in `completedMaintenanceOutcome`. `runBoundedReconcile` and the runtime-lock reaper now both follow that rule

internal/storage/executionstore/agent_wakeups_store.go
asherdale: agreed, ended up safely removing the rebuild path. every current production mutation path maintains the invariant transactionally: runnable work has either a wakeup or a runtime lock. integration tests now assert this after each current workflow

this is not a DB-level constraint, though, so a future write path could still break the invariant - the tests protect the existing paths, not arbitrary future code

internal/httpapi/apierror/errors.go
```diff
@@ -93,12 +93,22 @@ type ResponseError struct {
...
 
 func (err ResponseError) Error() string {
 	return err.Message
 }
 
+func (err ResponseError) Unwrap() error {
+	return err.cause
+}
+
+func (err ResponseError) WithCause(cause error) ResponseError {
```
ksarangmath, line 107: Some authorization paths currently drop the actual error and return Forbidden, which makes “the user was denied” indistinguishable from “we couldn’t perform the authorization check.” Can we preserve those as separate outcomes through the HTTP boundary, so a real denial becomes 403/404 while an operational failure keeps its cause and is handled as an operational error?

internal/log/http.go
```diff
@@ -18,6 +23,22 @@ func HTTPRequest(
...
+			fields := Fields{
+				"http.request.cancel_cause":  context.Cause(ctx).Error(),
+				"http.request.cancel_source": "request_context",
+			}
+			if deadline, ok := ctx.Deadline(); ok {
+				fields["http.request.deadline"] = deadline.UTC()
+			}
+			f.Attach(fields)
+			if errutil.OnlyMatches(f.e.err, context.Canceled) {
+				status = statusClientClosedRequest
```
ksarangmath, line 36: This changes the status to 499 only on the log event, while the metrics middleware still records the 500/503 written by the handler, so there might be mismatch between logs and metrics

internal/harness/worker/worker.go
ksarangmath: nit: errors.Is is also true when cancellation is joined with a real error, so we could drop a provider or persistence failure during shutdown.

ksarangmath (approved):
LGTM, just some minor comments

internal/httpapi/apierror/errors.go
asherdale: split operational errors from actual denials across the affected auth paths. `Authorize*` errors now preserve their cause and return an operational response

internal/harness/worker/worker.go
asherdale: shutdown suppression now requires the entire error tree to be cancellation-only. a joined cancellation plus provider/persistence failure remains visible at error level

internal/log/http.go
asherdale: good catch - logs and metrics now use the same final telemetry status. cancellation-only requests record 499 in both while keeping the wire response unchanged. genuine or mixed failures still record the original 5xx

PR #443, opened 2026-08-28
fix: Implement anthropic ephemeral caching by default

cmsparks:


api/openapi/openapi.yaml
```diff
@@ -6901,10 +6901,12 @@ components:
     ModelCacheRetention:
```
cmsparks, line 7023: at some point we probably want to deprecate / rework this field. i think it probably should be specific to the agent

internal/model/openaichatcompletions/request.go
```diff
@@ -65,9 +65,15 @@ func (p protocol) BuildRequest(ctx context.Context, input model.PrepareInput) (j
...
+	cacheRetention := model.EffectiveCacheRetention(
+		modelprotocol.APIFormatOpenAIChatCompletions,
+		apiVariant,
+		providerModelSlug,
+		input.Policy.CacheRetention,
+	)
 	if apiVariant == modelprotocol.APIVariantOpenRouter {
-		payload.CacheControl = openRouterCacheControl(input.Policy.CacheRetention, providerModelSlug)
-	} else if retention := promptCacheRetention(input.Policy.CacheRetention); retention != "" {
+		payload.CacheControl = openRouterCacheControl(cacheRetention, providerModelSlug)
```
ksarangmath: OpenRouter’s top-level cache_control only works when routing directly to Anthropic and excludes Bedrock and Vertex. Could we use explicit per-content-block cache_control markers instead? OpenRouter supports those across compatible upstream providers, and Pi v2, Goose, and Hermes use that approach.

internal/model/model.go
```diff
@@ -308,11 +308,37 @@ func containsModality(values []string, want string) bool {
...
+	}
+	if apiFormat == modelprotocol.APIFormatOpenAIChatCompletions &&
+		apiVariant == modelprotocol.APIVariantOpenRouter &&
+		isOpenRouterAnthropicSlug(providerModelSlug) {
+		return CacheRetentionShort
+	}
+	return CacheRetentionNone
+}
+
+func isOpenRouterAnthropicSlug(providerModelSlug string) bool {
```
ksarangmath: Could this share the existing openRouterSupportsTopLevelCacheControl decision instead of introducing a second anthropic/claude- classifier? These checks must remain identical: if they drift, we could resolve unset retention to short here but omit cache_control during serialization.

ksarangmath: nit: also could we just check for anthropic/ instead of anthropic/claude-, that's what the other harnesses do

ksarangmath (approved):
LGTM, just one main comment on the openrouter top level cache control because that eliminates some providers on openrouter

kvnloo:
CONTRIBUTING: one problem (Anthropic cache default). Conventional `fix:` title. Please confirm generated files (`api/openapi/openapi.yaml`, `frontend/packages/sdk/src/generated/*`, `internal/httpapi/openapi/api.gen.go`, `docs/api-reference/openapi.yaml`) came from `make openapi-generate && make docs-openapi && make web-generate` rather than hand edits.

Ties to #445 (prompt caching visibility/stability/duration). This PR is the default-on ephemeral cache path, not the full visibility FR — worth saying that in the PR body so #445 is not closed as duplicate.

Checkable from the file list: `internal/model/anthropicmessages/request.go` + `request_test.go`, `internal/modelprovider/resolver.go` + tests, `tools/ci/openapi-compat/approved-breaking-changes.txt`.

I have not merged-judgment the whole 891/674 yet. Ask: does default-on cache change billing/latency for existing agents without a grant flag? If yes, call that out next to the OpenAPI breaking-changes approval.

PR #444, opened 2026-08-29
fix: expose agent IDs and ignore Slack system messages

asherdale:
- adds each agent public ID to its model system prompt immediately after the default Omnara instructions
- ignores Slack system notifications from `USLACK` before inbound routing so channel-removal notices cannot create DM agents
- adds regression coverage for the prompt identity and the production-shaped Slack system event


ksarangmath (approved):
LGTM

PR #448, opened 2026-08-31
fix: deliver tool result images to chat completions models

### LINKED ISSUE #447: Inline media on custom tool results is stored as an artifact but never delivered to the model
Context: agent with custom tools, results posted via `POST /tool-calls/{id}/result`.
Reproduced on two vision-capable models (`deepseek/deepseek-v4-flash-vision-exp`,
`z-ai/glm-5.3-flash`) - same behavior both times.

**Steps**

1. Submit a tool result with an inline media block (a PNG screenshot). Accepted per the
   spec: *"Submission accepts inline media. The persisted response exposes that media as
   a `media_ref` block."*
2. The `tool_result` event carries the `media_ref` and the artifact is stored correctly
   (`image/png`, correct size, matching sha256).
3. Ask the model to describe the image.

**Actual**

The model never receives the image. In its context the tool result is text with this line
appended:

> A prior attachment with artifact ID art_… is not included in the current model context.

That string is Omnara's own, so the exclusion happens in model-context assembly, after storage.

**Expected**

The model receives the image. Attaching the identical media block to an agent input works -
the model describes it correctly. Usage agrees: a tool-result image adds ~0 input tokens to
the next model call, an input attachment adds normal image tokens.

**Impact**

Any agent whose tools produce images (screenshots, charts, rendered files) can never see its
own output. Only path today is the user re-attaching the file manually.

Happy to share agent/event/artifact IDs or a minimal repro.

**tldr**: Attaching images as user input works fine and models can see it, but if the agent comes across an image via a tool call, it can't see the actual image.

asherdale:
- Include custom-tool result images in Chat Completions media projection so selected artifacts remain available to the model.
- Keep all required `role: tool` messages contiguous, then attach resolved images in a follow-up multimodal user message with tool and artifact provenance.
- Preserve the existing textual fallback for omitted or unresolved media while avoiding duplicate artifact references for images that are rendered.
- Closes #447.

### DIFF
--- a/internal/model/openaichatcompletions/media.go
+++ b/internal/model/openaichatcompletions/media.go
@@ -20,7 +20,7 @@ func (p protocol) ProjectRenderedMedia(bundle modelcontext.Bundle) []modelcontex
 		)
 	}
 	for _, occurrence := range modelcontext.ResolvedMediaOccurrences(bundle) {
-		if occurrence.MessageRole != modelprotocol.RoleUser ||
+		if (occurrence.MessageRole != modelprotocol.RoleUser && !occurrence.IsToolResult()) ||
 			occurrence.Media.Kind != modelcontext.AttachmentKindImage {
 			continue
 		}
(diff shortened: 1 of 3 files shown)


internal/model/openaichatcompletions/media.go
```diff
@@ -113,8 +113,43 @@ func chatImagePart(resolved modelcontext.ResolvedMedia) (map[string]any, bool) {
...
+		if !ok {
+			continue
+		}
+		block, ok := chatImagePart(resolved)
+		if !ok {
+			continue
+		}
+		blocks = append(blocks, map[string]any{
+			"type": "text",
+			"text": "Attachment returned by tool " + result.Name + ". artifact_id: " +
```
ksarangmath: nit: Could we include the provider call ID in this attachment provenance text? Parallel calls can invoke the same tool more than once, and their images are all collected into this one user message. With only the tool name and artifact ID, it isn’t explicit which image belongs to which call. The provider call ID is already present in the assistant tool call and its tool-result message, so including it here would preserve that association.

ksarangmath (approved):
LGTM!

internal/model/openaichatcompletions/media.go
asherdale: good call, added the provider `tool_call_id` to each attachment label so parallel calls to the same tool remain unambiguous

PR #450, opened 2026-08-31
feat: add file attachments to agent chat

asherdale:
- add file picking, drag-and-drop, paste handling, previews, and attachment-only sending to the agent chat composer
- render sent and queued attachments with shared downloadable artifact cards while preserving optimistic attachment state
- expand and document validated image, text, and document ingestion, including Slack filename and MIME correction
- project attachments according to OpenAI Responses, chat-completions, OpenRouter, and Anthropic capabilities with UTF-8 text fallback and safe history degradation


frontend/apps/web/src/components/agents/AgentComposer.tsx
```diff
@@ -49,30 +211,78 @@ export function AgentComposer({
...
+          readOnly={submitting}
+          onChange={(event) => {
+            setText(event.target.value)
+          }}
+          onKeyDown={onKeyDown}
+          onPaste={(event) => {
+            const files = event.clipboardData.files
+            if (files.length === 0 || !ready || model == null || busyRef.current) return
+            void addFiles(files)
+          }}
```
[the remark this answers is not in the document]
asherdale: accepted file pastes now call `preventDefault()`, so clipboard text or file paths aren't inserted into the composer

frontend/apps/web/src/components/agents/AgentInputQueue.tsx
```diff
@@ -20,14 +20,15 @@ import {
...
 } from '@omnara/react'
-import { GripVertical, X } from 'lucide-react'
+import { File, GripVertical, X } from 'lucide-react'
 import { startTransition, useOptimistic } from 'react'
 
 import { errorMessage } from '@/lib/submit-status'
 
 function inputPreview(input: AgentInputBacklogItem) {
-  if (input.delivery_mode === 'optimistic') return input.text
+  if (input.delivery_mode === 'optimistic') return { text: input.text, attachmentCount: 0 }
```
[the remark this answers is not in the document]
asherdale: optimistic inputs now retain their attachment count after releasing the file payload

api/openapi/openapi.yaml
```diff
@@ -8856,6 +8856,15 @@ components:
...
             - text/tab-separated-values
+            - text/x-iif
+            - application/msword
+            - application/rtf
+            - application/vnd.oasis.opendocument.text
+            - application/vnd.apple.pages
+            - application/vnd.apple.keynote
+            - application/vnd.apple.iwork
+            - application/vnd.ms-powerpoint
+            - application/vnd.ms-excel
```
cmsparks, line 8970: originally, i set the content types which were commonly supported across responses / messages. technically, with completions some of these filetypes (word, etc) aren't supported. 

For unsupported filetypes, do we just fall back to just passing the artifact_id in, and then the agent can interact with the file after downloading it? (we probably always want to include artifact id though fwiw)

internal/model/openaichatcompletions/media.go
```diff
@@ -105,12 +135,42 @@ func userChatContentFromParts(raw json.RawMessage, media map[string]modelcontext
...
+}
+
+func rendersAsChatText(resolved modelcontext.ResolvedMedia) bool {
+	return modelcontext.IsTextDocumentMediaType(resolved.MediaType) &&
+		(len(resolved.Data) == 0 || utf8.Valid(resolved.Data))
+}
+
+func rendersAsChatDocument(resolved modelcontext.ResolvedMedia) bool {
+	return rendersAsChatText(resolved) || resolved.MediaType == "application/pdf"
 }
```
cmsparks, line 177: preexisting issue, but for completions we probably want to enable the openrouter PDF plugin by default: https://openrouter.ai/docs/guides/features/plugins

api/openapi/openapi.yaml
```diff
@@ -8866,6 +8875,7 @@ components:
           type: string
           format: byte
           minLength: 1
+          description: Base64-encoded file bytes. For text media types, the decoded bytes must be valid UTF-8.
```
cmsparks, line 8981: we should probably include more documentation on when certain filetypes are accepted and when they get sent to the model vs as artifact ids

cmsparks (approved):

api/openapi/openapi.yaml
asherdale: good call, unsupported documents now fall back to the filename + artifact ID, and artifact IDs are included alongside files sent directly

internal/model/openaichatcompletions/media.go
asherdale: yessir good call on this too. after looking into it, it looks like PDF inputs are already implicitly enabled for every OpenRouter request containing a PDF, and our workspace default is already Mistral OCR: https://openrouter.ai/docs/guides/overview/multimodal/pdfs

so it looks like we should be good on this front already, but if i've misunderstood something feel free to lmk and i'd be happy to add an explicit override here

frontend/apps/web/src/lib/agent-attachments.ts
```diff
@@ -0,0 +1,157 @@
...
+  mediaType: AttachmentMediaType,
+): boolean {
+  if (kind === 'image') return allowsModality(model, 'image')
+  if (mediaType.startsWith('text/')) return allowsModality(model, 'text')
+  if (model.api_format === 'openai-chat-completions' && mediaType === 'application/pdf') {
+    return model.api_variant === 'openrouter' || allowsModality(model, 'file')
+  }
+  if (model.api_format === 'openai-responses' || mediaType === 'application/pdf') {
+    return allowsModality(model, 'file')
+  }
```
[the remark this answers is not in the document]
asherdale: this is intentional — `text,image` excludes file input, and the backend enforces the same rule. empty modalities or `file` allows Anthropic PDFs

PR #451, opened 2026-08-31
fix: preserve inherited tool permissions while catalogs load

asherdale:
- ignore transient empty permission values emitted while tool catalogs and MCP permission profiles load
- preserve inherited permissions instead of writing invalid empty overrides into the agent draft
- add regressions covering both asynchronous permission-loading paths

relevant upstream Radix issue: https://github.com/radix-ui/primitives/issues/3135


ishaansehgal99 (approved):
lgtm

PR #452, opened 2026-08-31
feat: Add skill editing and viewing

cmsparks:


internal/httpapi/skill_routes.go
```diff
@@ -468,9 +622,45 @@ func (s strictOpenAPIServer) GetSkill(
 	if err != nil {
 		return nil, err
 	}
+	files, err := s.loadSkillFiles(ctx, record)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
cmsparks: yes, this is intended. failing to load the skill should fail api.

frontend/apps/web/src/components/skills/UpdateSkillDialog.tsx
```diff
@@ -0,0 +1,202 @@
...
+                </Field>
+              </TabsContent>
+              <TabsContent value="archive">
+                <Field>
+                  <FieldLabel htmlFor={archiveInputId}>Skill archive</FieldLabel>
+                  <div className="border-border bg-muted/20 flex items-center gap-3 rounded-lg border border-dashed p-4">
+                    <div className="bg-background flex size-9 shrink-0 items-center justify-center rounded-md border">
+                      <FileArchive className="text-muted-foreground size-4" />
+                    </div>
+                    <Input
```
asherdale: minor nit: could we style `Choose File` as a distinct element and the filename/status as separate text? right now they look pretty similar, almost like a continuous label:

Image

Image

so this would apply to the 'upload skill' dialog too

frontend/apps/web/src/components/skills/SkillMdEditor.tsx
```diff
@@ -0,0 +1,93 @@
...
+  id: string
+  defaultValue: string
+  onChange: (value: string) => void
+  readOnly?: boolean
+  className?: string
+}) {
+  const editorElementRef = useRef<HTMLDivElement | null>(null)
+  const editorRef = useRef<monaco.editor.IStandaloneCodeEditor | null>(null)
+  const emitChange = useEffectEvent(onChange)
+  const initialValueRef = useRef(defaultValue)
```
asherdale: should we sync prop updates into Monaco? a refetch can update `currentMd`, but `initialValueRef` leaves stale content in the editor

internal/httpapi/skill_routes.go
```diff
@@ -240,6 +232,174 @@ func readSkillUpload(body *multipart.Reader, principal identitystore.PrincipalRe
...
+	} else {
+		record, recordErr := s.server.skills.GetVisibleSkill(ctx, org.ID, request.SkillID, principal)
+		if recordErr != nil {
+			return nil, skillAPIError(ctx, recordErr)
+		}
+		revisionID, revisionErr := publicID(publicid.KindSkillRevision, record.RevisionID)
+		if revisionErr != nil {
+			return nil, revisionErr
+		}
+		current, _, loadErr := s.server.skills.LoadSkillArchive(ctx, request.SkillID, revisionID)
```
asherdale, line 360: can we reject this if `revisionID` is no longer current? otherwise a concurrent update can be replaced with stale archive files

asherdale (approved):
lgtm! would also fix the upload limit bugbot comment before merging

frontend/apps/web/src/components/skills/SkillMdEditor.tsx
cmsparks: yeah if they click off, i'll do that (though we shouldn't regularly refetch while the user has the dialog open)

PR #464, opened 2026-09-01
feat: Automatic login in CLI when running command

cmsparks:
When running a command, we should automatically login


asherdale (approved):
lgtm!

PR #465, opened 2026-09-01
feat: support sending artifacts to Slack

asherdale:
- allow send_integration_message to accept an optional artifact_id and load the referenced artifact within the active agent and project
- implement the Slack external file upload flow with signed URL validation, runtime ownership checks before each request, and phase-aware retry behavior
- request files:write for Slack installs and return actionable reconnect guidance when an existing installation lacks the scope
- update the built-in tool schema and documentation for artifact delivery
image

### DIFF
--- a/internal/harness/tools/integration_tools.go
+++ b/internal/harness/tools/integration_tools.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/omnara-ai/omnara/internal/integration/slack"
 	"github.com/omnara-ai/omnara/internal/model"
+	"github.com/omnara-ai/omnara/internal/modelcontext"
 	"github.com/omnara-ai/omnara/internal/publicid"
 	"github.com/omnara-ai/omnara/internal/storage"
 	"github.com/omnara-ai/omnara/internal/storage/executionstore"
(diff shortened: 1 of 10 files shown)


internal/toolcatalog/catalog.go
```diff
@@ -360,6 +361,10 @@ func integrationSendTool() (Entry, error) {
 				"minLength":   1,
 				"description": "User-visible message text to send to the current integration target.",
 			},
+			"artifact_id": map[string]any{
```
cmsparks: should we allow multiple artifact ids instead of just one? my gut feel is that we probably want this to be an array, and then the integration might define whether it's sent as one or multiple messages

internal/harness/tools/integration_tools.go
```diff
@@ -252,6 +237,129 @@ func (e Executor) dispatchIntegrationMessageSend(
...
+		artifactID,
+	)
+	if err != nil {
+		return toolResultContent{}, fmt.Errorf("load artifact %s: %w", input.ArtifactID, err)
+	}
+	filename := modelcontext.MediaFilename(artifact.Filename, artifact.ContentType)
+	var rateLimitSlept time.Duration
+	attempt := 1
+	var fileID string
+	for {
```
cmsparks: nit: this is a little confusing that we have a infinite loop where it's controlled by return / break / continue (same comment applies below as well)

I'd prefer if we had a well defined loop for the retries (i.e. for (attempt := 0; i < maxAttempts; attempt++))

cmsparks (approved):
lgtm

internal/toolcatalog/catalog.go
asherdale: yea great call here - now this new arg is an array

PR #471, opened 2026-09-02
feat: Add steering / queued control to cron triggers

cmsparks:


api/openapi/openapi.yaml
```diff
@@ -8666,6 +8677,8 @@ components:
           $ref: '#/components/schemas/CronTimezone'
         message_template:
           $ref: '#/components/schemas/CronMessageTemplate'
+        delivery_mode:
```
ksarangmath: Should this go inside CronTriggerTarget, it seems better suited there because it describes the type of the target, and delivery mode only makes sense if target is type agent

api/openapi/openapi.yaml
```diff
@@ -8638,6 +8638,17 @@ components:
...
         render is recorded in `failure_report` without sending a message.
       maxLength: 65536
+    CronTriggerDeliveryMode:
+      type: string
+      enum: [queued, steering]
+      default: queued
+      description: >-
+        How each firing's message is delivered to an existing agent target. `queued` appends the
+        message to the agent's input backlog, so firings accumulate while the agent is busy.
+        `steering` delivers the message into the agent's current turn, or opens a new turn when
```
ksarangmath: nit: technically steering messages create a new turn, not add to an existing turn. but more broadly, maybe we just say it fires a steering message, and we don't have to explain steering here, it should be explained on agent input

internal/storage/executionstore/cron_triggers_store.go
```diff
@@ -36,6 +36,7 @@ type CreateCronTriggerInput struct {
 	CronExpression  string
 	Timezone        string
 	MessageTemplate string
+	DeliveryMode    AgentInputDeliveryMode
```
ksarangmath: nit: AgentInputDeliveryMode includes 'immediate', which shouldn't be allowed for cron delivery mode

frontend/apps/web/src/components/agents/cron-trigger-delivery-mode.ts
```diff
@@ -0,0 +1,28 @@
...
+export const cronTriggerDeliveryModeOptions: CronTriggerDeliveryModeOption[] = [
+  {
+    value: 'queued',
+    label: 'Queue',
+    hint: "Each firing waits in the agent's backlog until earlier inputs finish.",
+  },
+  {
+    value: 'steering',
+    label: 'Steer',
+    hint: 'Each firing is delivered into the current turn, or starts a new turn if the agent is idle.',
```
ksarangmath: nit: same point as above about it starting a new turn, but since this is user facing and technically this is easier to understand, I'm fine with this, or maybe some slight change

frontend/apps/web/src/components/agents/CronTriggersSection.tsx
```diff
@@ -83,11 +84,12 @@ export function CronTriggersList({
                       {trigger.message_template}
                     </TooltipContent>
                   </Tooltip>
-                  {!trigger.enabled && <Badge variant="secondary">Disabled</Badge>}
```
ksarangmath, line 86: nit: This badge was visible regardless of canManage, while the replacement pause/play control is inside the canManage block below. Removing it means read-only users no longer see that a schedule is disabled.

internal/crontrigger/service.go
```diff
@@ -189,7 +189,7 @@ func (s *Service) sendAgentInput(
 		AgentID:        trigger.Target.ID,
 		Actor:          actor,
 		ContentBlocks:  contentBlocks,
-		DeliveryMode:   executionstore.DeliveryModeQueued,
+		DeliveryMode:   trigger.DeliveryMode,
```
ksarangmath: Not directly related to this change, and it's an edge case that only happens when a trigger config is changed, but since the created agent input is the evidence that this occurrence was delivered, is it possible to create it and complete the cron firing in the same transaction? Currently the input can commit and the worker can die before CompleteCronTriggerFiring, causing an already-delivered occurrence to be retried. Changing the template or delivery mode before that retry then turns it into an idempotency conflict.

ksarangmath (approved):
LGTM

api/openapi/openapi.yaml
cmsparks: That's a very good point, i'll update to do that

frontend/apps/web/src/components/agents/CronTriggersSection.tsx
cmsparks: set them to disabled (greyed out) on !can manage

PR #472, opened 2026-09-02
fix: Fix frontend question UI

cmsparks:


frontend/apps/web/src/components/agents/AgentInteractions.tsx
```diff
@@ -34,6 +35,9 @@ function keyedContextItems(interaction: AgentInteraction) {
   })
 }
 
+// Matches the title produced by internal/harness/tools/permission_modes.go.
+const permissionTitlePrefix = 'Permission requested for '
```
ksarangmath: The server should have the tool call for a permission interaaction, I think it might just not be exposed in the API. If that is a viable path, we could remove this string check and use the tool call directly from what the server stores.

frontend/apps/web/src/components/agents/AgentInteractions.tsx
```diff
@@ -110,16 +127,24 @@ function InteractionFormCard({
           const allowsText = selectedOptions(question.options, optionIndices).some(
             (option) => option.allows_text === true,
           )
+          const isLast = questionIndex === questions.length - 1
+          const responseId = `${interaction.id}-${questionIndex}-response`
+          const responseLabel = isPermission ? 'Reason (optional)' : 'Details (optional)'
```
ksarangmath, line 128: docs/events/interactions.mdx may need to be updated with new behavior

ksarangmath (approved):
LGTM, left one potentially straightforward thing where we could use server side info rather than string parsing, but I didn't look too deeply into it

frontend/apps/web/src/routes/AgentView.tsx
```diff
@@ -221,3 +183,73 @@ export function AgentView() {
...
+    <div className="mx-auto grid w-full max-w-3xl shrink-0 gap-3 pt-3">
+      {!archived && !configOpen && (
+        <AgentInteractions
+          orgID={orgID}
+          projectID={projectID}
+          agentID={agentID}
+          agentActive={chat.isWorking}
+          canOperate={canOperate}
+        />
+      )}
```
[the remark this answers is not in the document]
cmsparks: this is intended.

PR #473, opened 2026-09-02
chore: Antislop

cmsparks:
* add antislop eslint tools (in rules/), these are copied from https://github.com/dmmulroy/anti-slop/tree/main
* fix all linting issues
* use oxlint instead of eslint (essentially faster eslint)


frontend/package.json
ksarangmath: pnpm lint failed on my machine because I have node 22.17, for pnpm lint we should probably bump this up to be >= 24. CI is also on node 24

frontend/.oxlintrc.json
```diff
@@ -0,0 +1,510 @@
...
+    ".claude/**",
+    ".codex/**",
+    ".continue/**",
+    ".cursor/**",
+    ".gemini/**",
+    ".opencode/**",
+    ".pi/**",
+    ".roo/**",
+    ".windsurf/**",
+    "tools/oxlint/anti-slop/**"
```
ksarangmath, line 18: nit: since this is our codebase now, should we remove this? or if it's not planning to ever be edited, then I guess we can keep

ksarangmath: nit: I think pretty much all of this also is covered by gitignore, except the sdk generated path

frontend/.oxlintrc.json
```diff
@@ -0,0 +1,510 @@
...
+  ],
+  "rules": {
```
ksarangmath, line 20: nit: Should use oxlint defaults rather than hardcoding this list? Looks like they have categories that we can enable, "correctness" being the main one

also just for info: no-undef and jsx-a11y/lang were in eslint previously and are in oxlint, but not enabled here
no-octal, no-dupe-args, and react-hooks 6.1's component-hook-factories / config / gating are not in oxlint at all but were in eslint, just mentioning in case there's equivalents or something that we'd wanna keep.

frontend/.oxlintrc.json
```diff
@@ -0,0 +1,510 @@
...
+    "anti-slop/no-chained-type-assertions": "error",
+    "anti-slop/no-conditional-empty-object-spread": "error",
+    "anti-slop/no-known-value-widening": "error",
+    "anti-slop/no-module-mocking": "error",
+    "anti-slop/no-object-parameters": "error",
+    "anti-slop/no-reflect-apply": "error",
+    "anti-slop/no-reflect-get": "error",
+    "anti-slop/no-runtime-typeof": "error",
+    "anti-slop/no-shape-in-symbol-names": "error",
+    "anti-slop/no-unknown-parameters": "error",
```
ksarangmath, line 225: This rule causes some smells (same with no-unknown-returns and no-unknown-type-aliases below), using unknown is sometimes cleaner than not, specifically with data that's coming from external sources. Maybe the rule can be limited to things that don't get data from external sources, or we make it a warning?

frontend/packages/sdk/src/agent-event-stream.ts
```diff
@@ -231,16 +234,21 @@ async function* openConnection(
   }
 }
 
+function errorFromBody(status: number, body: { error?: unknown }['error']): ApiError {
```
ksarangmath: nit: this is kinda hiding from the above rule by making it roundabout way of using unknown

{ error?: unknown }['error']

frontend/packages/sdk/src/agent-event-stream.ts
```diff
@@ -231,16 +234,21 @@ async function* openConnection(
...
+  const json = zJsonBody.safeParse(body)
+  return ApiError.fromBody(status, json.success ? json.data : undefined)
+}
+
 function decodeFrame(text: string): AgentEventStreamData {
-  let data: unknown
+  let data: JsonBody
   try {
-    data = JSON.parse(text)
+    data = zJsonBody.parse(JSON.parse(text))
```
ksarangmath: nit: same as above, I don't think this check is needed cuz it was checked later, and it's only doing this so it's not unknown

frontend/packages/sdk/src/errors.ts
```diff
@@ -1,24 +1,30 @@
...
-import { zAgentConfigErrorIssue } from './generated/zod.gen'
+import { zAgentConfigErrorIssue, zError } from './generated/zod.gen'
+import { type JsonBody, jsonBody } from './json-body'
 
 export type ApiErrorCode = ApiErrorBody['code']
 
 const errorIssuesSchema = z.object({ issues: z.array(zAgentConfigErrorIssue).optional() })
 
-function issuesFromBody(body: unknown): AgentConfigErrorIssue[] {
+const errorResponseSchema = zError
```
ksarangmath: nit: this could cause new error codes from the server to fail and not show the message

ksarangmath: Maybe using relaxedSchema here can solve the issue

frontend/.oxlintrc.json
```diff
@@ -0,0 +1,510 @@
...
+    ],
+    "typescript/prefer-optional-chain": "error",
+    "anti-slop/no-chained-type-assertions": "error",
+    "anti-slop/no-conditional-empty-object-spread": "error",
+    "anti-slop/no-known-value-widening": "error",
+    "anti-slop/no-module-mocking": "error",
+    "anti-slop/no-object-parameters": "error",
+    "anti-slop/no-reflect-apply": "error",
+    "anti-slop/no-reflect-get": "error",
+    "anti-slop/no-runtime-typeof": "error",
```
ksarangmath: nit: This rule forces some good changes, e.g. in the monacoEnvironment.ts file, but also causes some smells. Allowing it in typeguards and only setting it in boundaries to external stuff could make things cleaner

packages/sdk/src/auth.ts:6
useAgentBuilderForm.ts:405
places where we use instanceof is used over === and makes the statement weaker

validate-response.ts
flags.ts
output.ts
places where we use zod instead of using type of

frontend/packages/sdk/src/validate-response.ts
```diff
@@ -1,59 +1,61 @@
...
-  schema: T,
-  data: unknown,
-): z.ZodSafeParseResult<z.output<T>> {
-  const result = schema.safeParse(data)
-  if (result.success || isUnknownEnumError(result.error, data)) {
-    return { success: true, data: data as z.output<T> }
-  }
-  return { success: false, error: result.error }
+export function relaxedSchema<S extends z.ZodType>(schema: S): z.ZodType<z.output<S>> {
+  return z.custom<z.output<S>>((value) => {
```
ksarangmath: nit: z.custom drops the inner ZodError, so on a real failure this just says "Invalid input" instead of which field was wrong. before it returned the actual error, so decodeFrame's cause in agent-event-stream.ts used to be useful.

frontend/.oxlintrc.json
```diff
@@ -0,0 +1,510 @@
...
+    "anti-slop/no-shape-in-symbol-names": "error",
```
ksarangmath, line 224: nit: no-shape-in-symbol-names bans the substring "shape" in every identifier, including property access, so zod's schema.shape is unwritable here (the agent-attachments.test.ts rewrite is that). Pointing out cuz it might get worse the more we use zod

frontend/.oxlintrc.json
```diff
@@ -0,0 +1,510 @@
...
+    "typescript/no-unnecessary-condition": [
+      "error",
+      {
+        "allowConstantLoopConditions": "only-allowed-literals"
+      }
+    ],
+    "typescript/prefer-optional-chain": "error",
+    "anti-slop/no-chained-type-assertions": "error",
+    "anti-slop/no-conditional-empty-object-spread": "error",
+    "anti-slop/no-known-value-widening": "error",
```
ksarangmath, line 213: nit: PR now has three ways of writing a lookup table (satisfies, Map, and an if/else at AgentProfileIntegrations.tsx:12). If the rule stays, worth stating the intended one here: known keys → satisfies Record<K, V>, arbitrary keys → Map. noUncheckedIndexedAccess is on in the compiler, so this rule might not being doing much anways

frontend/apps/web/src/test/fake-api.ts
```diff
@@ -0,0 +1,85 @@
...
+export function neverResponds(): Promise<Response> {
+  return new Promise(() => undefined)
+}
+
+function requestUrl(input: RequestInfo | URL): URL {
+  const href = input instanceof Request ? input.url : String(input)
+  return new URL(href, window.location.href)
+}
+
+function requestBody(init: RequestInit | undefined): JsonValue | undefined {
```
ksarangmath: This reads the body from init, but the SDK client calls fetch(request) with a Request object and no init, so every SDK-driven request gets recorded with body: undefined

ksarangmath: Could normalize it first
const request = input instanceof Request ? input.clone() : new Request(input, init)
and then read from request

frontend/apps/web/src/components/org/ProjectAccessEditor.tsx
```diff
@@ -47,71 +226,44 @@ export function ProjectAccessEditor({
...
-            if (projectsQuery.isError) void projectsQuery.refetch()
-          }}
-        >
-          Retry
-        </Button>
-      </div>
+      <LoadError
+        message={errorMessage(accessQuery.error, 'Could not load project access.')}
+        retrying={accessQuery.isFetching}
+        onRetry={accessQuery.refetch}
```
ksarangmath: nit: this passes the click event into refetch(options). Line 254 has the right form: () => void accessQuery.refetch()

ksarangmath (approved):
mostly a bunch of straightforward nits, I think some of the rules might be making the code smelly in places, but overall looks good

frontend/.oxlintrc.json
cmsparks: IMO unknown is pretty much always an antipattern (and is the main thing I want to get out of this PR), namely causing things like isRecord or overly aggressive typecasting.

though there definitely are some issues where it's being hidden (namely the error?: unknown below and json body issue) though there are ways to get around the intermediate unknown variable,

frontend/packages/sdk/src/agent-event-stream.ts
cmsparks: i'll rewrite this so that it's not doing this

frontend/.oxlintrc.json
cmsparks: these generally won't ever change, so i'll probably leave it excluded for now, but i'll update to remove the gitignored files

frontend/.oxlintrc.json
cmsparks: set it to allowIntypeguards and then in other places just replaced with zod schemas

frontend/.oxlintrc.json
cmsparks: ah this is actually a bug in the linter upstream, should be allowed now

frontend/apps/web/src/test/fake-api.ts
cmsparks: updated this to fix

frontend/.oxlintrc.json
cmsparks: normalized all of this

frontend/packages/sdk/src/validate-response.ts
cmsparks: fixed this, wrapped in the relaxed schema

frontend/packages/sdk/src/agent-event-stream.ts
cmsparks: fixed this, used some clearer zod schemas here

frontend/.oxlintrc.json
cmsparks: added the rules we were missing and added correctness rules

PR #474, opened 2026-09-03
feat: Subagents

cmsparks:
Implement subagents.

Configuring subagents:
```yaml
# new top level subagents field
subagents:
  researcher:
    # type profile references another agent profile as the config
    type: profile
    profile: research-agent
    description: Read-only codebase investigation.
    max_concurrent: 4
  fork:
    # type self reuses the current agent profile and machines, but with a clear context
    # it omits subagent tools to prevent infinite recursion
    type: self
    instruction:
      append: You are a subagent. Finish with a concise summary.
```

Tool schema:
* spawn_agent(key, task, name): start a subagent
* read_agent(agent_ref, after_sequence, before_sequence, limit): read subagent messages
* list_agents(): list active agents
* send_agent_message(agent_ref, message): send an agent a message to steer
* stop_agent(agent_ref): archive agent

Upon an agent stopping, it will automatically send a message to the parent agent with a notification and excerpt of the existing message. This gets linked with the external actor id of the agent


migrations/000031_subagents.sql
```diff
@@ -0,0 +1,84 @@
...
+
+DROP TRIGGER agents_identity_immutable ON agents;
+
+CREATE TRIGGER agents_identity_immutable
+BEFORE UPDATE OF id, org_id, project_id, agent_profile_id, parent_agent_id,
+    subagent_key, idempotency_key, created_at ON agents
+FOR EACH ROW EXECUTE FUNCTION agents_reject_identity_change();
+
+ALTER TABLE tool_calls
+    ADD COLUMN deadline_at timestamptz;
```
cmsparks: currently, this is just for agent_wait, but we probably want to extend this to process functionality

internal/storage/executionstore/subagents_store.go
```diff
@@ -0,0 +1,790 @@
...
+	SubagentStateArchived       = "archived"
+)
+
+type SubagentLaunch struct {
+	ParentAgentID           ID
+	Key                     string
+	MaxConcurrent           *int
+	MaxSubagents            *int
+	ShareParentMachines     bool
+	ArchiveAfterIdleMinutes *int
```
ksarangmath: I was recently thinking about adding an archive timeout for even regular top level agents, should we do that in this PR? If not in this PR, I think we can just make sure that we're not pigeonholing ArchiveAfterIdleMinutes to being a subagent specific thing (which it doesn't seem like it from my review, schema looks good and everything for it to be generally used for top level agents too)

ksarangmath: Probably can do this in a follow up PR, unless you think otherwise

internal/storage/executionstore/subagents_integration_test.go
```diff
@@ -0,0 +1,542 @@
...
+	}
+	parent := parentLaunch.Agent
+
+	child, err := spawnSubagentForTest(
+		t, ctx, store, parent, profile.CurrentConfigID, "worker-1", "subagent-launch-child-1", intPtrForSubagentTest(1),
+	)
+	if err != nil {
+		t.Fatalf("spawn subagent: %v", err)
+	}
+	if child.Agent.ParentAgentID != parent.ID || child.Agent.SubagentKey != "fork" {
```
ksarangmath, line 118: nit: could we also add test for rejecting a parent from another project, and checking the last allowed depth agent and first rejected depth agent

internal/harness/tools/subagent_tools.go
```diff
@@ -0,0 +1,569 @@
...
+		[]byte(baseConfig.Source),
+	)
+	if err != nil {
+		return storage.NilID, storage.NilID, fmt.Errorf("parse base agent config source: %w", err)
+	}
+	childSource, err := json.Marshal(agentconfig.SubagentSource(baseSource, subagent))
+	if err != nil {
+		return storage.NilID, storage.NilID, fmt.Errorf("encode subagent config source: %w", err)
+	}
+	body, err := agentconfigcompile.Compile(
```
ksarangmath: Could we derive the child from baseConfig.CompiledDefinition instead of recompiling the saved source? The base config already resolved its model, machine, and pool names to IDs. Looking those names up again can fail after a rename, or select a different resource if a name was reused.

internal/agentconfig/runtime_contract.go
```diff
@@ -126,10 +134,23 @@ func RuntimeContractFromCompiled(
...
 	if len(compiled.Skills) > 0 {
-		return contract.WithImplicitBuiltInTool(toolcatalog.ToolNameSkill)
+		contract, err = contract.WithImplicitBuiltInTool(toolcatalog.ToolNameSkill)
+		if err != nil {
+			return RuntimeContract{}, err
+		}
+	}
+	if len(compiled.Subagents) > 0 {
+		for _, name := range toolcatalog.SubagentToolNames() {
+			contract, err = contract.WithImplicitBuiltInTool(name)
```
ksarangmath, line 155: nit: I forget how it's handled for the other implicit tools, but I think deleting the spawn_agent tool in the UI will implicitly make it always allow, maybe in the UI it shouldn't be able to be deleted?

internal/agentconfig/subagents.go
```diff
@@ -0,0 +1,242 @@
...
+func compileSubagents(
+	source AgentConfigSource,
+	opts CompileOptions,
+) (map[string]SubagentCompiled, error) {
+	keys := slices.Sorted(maps.Keys(source.Subagents))
+	compiled := make(map[string]SubagentCompiled, len(keys))
+	for _, key := range keys {
+		entry := source.Subagents[key]
+		pointer := jsonPointer("subagents", key)
+		if toolcatalog.IsSubagentToolName(key) {
```
ksarangmath: nit: is this restriction necessary?

internal/harness/tools/subagent_tools.go
```diff
@@ -0,0 +1,569 @@
...
+	}
+	actor, err := executionstore.SubagentActorParams(parent.OrgID, parent)
+	if err != nil {
+		return nil, err
+	}
+	name := input.Name
+	if name == "" {
+		name = input.Agent
+	}
+	launch, err := executor.Store.Execution().LaunchAgent(ctx, executionstore.LaunchAgentInput{
```
ksarangmath: Could we create the child and record the spawn_agent result in the same transaction?

internal/storage/executionstore/subagents_store.go
```diff
@@ -0,0 +1,790 @@
...
+	ctx context.Context,
+	tx pgx.Tx,
+	qtx *dbsqlc.Queries,
+	projectID ID,
+	launch SubagentLaunch,
+) (AgentRecord, error) {
+	if isNilID(launch.ParentAgentID) || launch.Key == "" {
+		return AgentRecord{}, errors.New("subagent launch requires a parent agent and key")
+	}
+	if _, err := qtx.LockAgentInProject(
```
ksarangmath: We can use the helpers from PR #300 to have the same lock ordering

internal/storage/executionstore/model_outputs_terminal_error.go
```diff
@@ -258,6 +261,16 @@ func publishModelCallErrorOutputTx(
 	}); err != nil {
 		return TypedAgentEventRecord{}, fmt.Errorf("reconcile wakeup after model call error: %w", err)
 	}
+	message := subagentMessage{
+		Kind:           SubagentMessageKindFailed,
+		Text:           input.ErrorMessage,
+		IdempotencyKey: "model_output:" + modelOutput.ID.String(),
+	}
+	if err := handleSubagentTurnEndedTx(
```
ksarangmath, line 269: Could we have recovery and archival acquire their parent/child locks through #300’s helpers, while keeping the reaper’s ability to skip busy agents? Recovery currently locks the child before reaching this parent notification, which can deadlock with archival.

internal/storage/executionstore/subagents_store.go
```diff
@@ -0,0 +1,790 @@
...
+) ([]MachineRecord, int, error) {
+	var machines []MachineRecord
+	archived := 0
+	for _, candidate := range candidates {
+		txNotifications := s.newTxNotifications()
+		tx, err := s.pool.BeginTx(ctx, pgx.TxOptions{})
+		if err != nil {
+			return machines, archived, fmt.Errorf("begin %s: %w", commitScope, err)
+		}
+		released, err := archiveAgentTreeTx(
```
ksarangmath: Should we recheck idle eligibility after acquiring the lifecycle locks, before canceling or archiving anything?

ksarangmath (commented):
LGTM, there's just one more part of the PR I need to finish reviewing, but I'll approve after that, here's some comments in the meantime

internal/storage/executionstore/subagents_store.go
cmsparks: fwiw yeah there's no reason we couldn't just have this be a general agent setting, I'll make it so it's not subagent specific, but I'll probably expose it on top level agents as a follow up

internal/agentconfig/runtime_contract.go
cmsparks: I think the same is the case with Skill(), I think I'll fix this as a followup (i was meaning to do some cleanup of the machine tools as well)

internal/harness/tools/subagent_tools.go
```diff
@@ -0,0 +1,573 @@
...
+		parent.OrgID,
+		parent.ProjectID,
+		baseConfig,
+		subagent,
+	)
+	if err != nil {
+		return subagentLaunchConfig{}, fmt.Errorf("derive subagent config: %w", err)
+	}
+	derived := body.CreateInput(parent.ProjectID)
+	return subagentLaunchConfig{derived: &derived}, nil
```
ksarangmath: Should we consistently treat agent_profile_id as attribution to the profile used as a template? Instruction/model overrides or passed-in machines could still keep that attribution, and for 'self', we could carry forward the parent’s source profile when present.

Basically slightly changing agent_profile_id to mean what profile was used to launch from, even though the actual config isn't from the agent profile.

We'd have to update the launch validation to support that meaning, it currently requires the effective config to belong to the profile’s version history, which wouldn’t hold for a derived config.

right now it's inconsistent because 'self' doesn't set agent_profile_id, and 'profile' with instruction  or model override doesn't set agent_profile_id, but 'profile' with no override does set agent_profile_id.

internal/agentconfigcompile/compile.go
```diff
@@ -0,0 +1,312 @@
...
+	if err != nil {
+		return Body{}, err
+	}
+	configuredModelID, err := storage.ParseID(child.Model.ConfiguredModelID)
+	if err != nil || configuredModelID == storage.NilID {
+		return Body{}, fmt.Errorf("subagent model must resolve to a configured project-granted model")
+	}
+	return Body{
+		Definition:         json.RawMessage(encoded.CanonicalJSON),
+		Source:             base.Source,
```
ksarangmath: This might not include overriden instruction when someone opens a subagent yaml, not sure what the best way to handle this is...

internal/harness/tools/dispatch.go
```diff
@@ -63,6 +63,7 @@ func (e Executor) dispatchToolHandler(
 					ctx,
 					handler.Transactional,
 					transactionalToolContext{
+						Executor:   e,
```
ksarangmath: Could we keep Executor out of the transactional context and add the reads needed by spawn_agent to ToolCallReader instead? Spawning already holds a database connection, but runtimeContractForTurn and the other global store calls request another, so spawning can stall when the pool is exhausted. Using call.Reader keeps those reads in the existing transaction. Could we also extend TestTransactionalToolDispatchUsesOneDatabaseConnection to cover spawn_agent?

internal/storage/executionstore/agent_launch_store.go
```diff
@@ -157,6 +199,28 @@ func (s *Store) launchAgentOnce(
...
 	}
+	var sharedBindings []dbsqlc.ListParentMachineBindingsForSharingRow
+	if input.Subagent != nil {
+		sharedBindings, err = admitSubagentLaunchTx(ctx, tx, qtx, project.OrgID, input.ProjectID, *input.Subagent)
+		if err != nil {
+			return LaunchAgentResult{}, err
+		}
+		insertInput.ParentAgentID = input.Subagent.ParentAgentID
+		insertInput.SubagentKey = input.Subagent.Key
+		agent, inserted, err = insertAdmittedAgentTx(ctx, tx, qtx, insertInput)
```
ksarangmath: Could we consolidate the insertion step so regular and subagent launches acquire their locks in the same order? Currently, regular launches acquire the agent quota before the machine pools, while subagent launches do the reverse, which can deadlock.

internal/agentconfig/subagents.go
```diff
@@ -0,0 +1,236 @@
...
+	SubagentTypeSelf    = "self"
+)
+
+type AgentConfigSubagentSource struct {
+	Type                    string                                `json:"type"`
+	Profile                 string                                `json:"profile,omitempty"`
+	Description             string                                `json:"description,omitempty"`
+	Model                   *AgentConfigSubagentModelSource       `json:"model,omitempty"`
+	Instruction             *AgentConfigSubagentInstructionSource `json:"instruction,omitempty"`
+	MaxConcurrent           *int                                  `json:"max_concurrent,omitempty"`
```
ksarangmath: very nit: should we call this MaxInstances, MaxConcurrent might come across like max number of agents that can be working at once

internal/harness/tools/subagent_tools.go
```diff
@@ -0,0 +1,573 @@
...
+		"key":       key,
+		"state":     executionstore.SubagentStateRunning,
+		"message": "Subagent started. Its final answer will arrive as a message from it; " +
+			"use read_agent to check its progress.",
+	})
+}
+
+func provisionSubagentMachinesInBackground(ctx context.Context, call backgroundToolContext) error {
+	launch, ok := call.CommandResult.(executionstore.LaunchAgentResult)
+	if !ok || len(launch.ProvisionMachineIDs) == 0 || call.Executor.MachinePoolManager == nil {
```
ksarangmath: nit: Could we return an error when CommandResult has the wrong type, separately from the valid case where there are no machines to provision? Otherwise, a wiring mistake silently skips provisioning and looks successful. The same applies to the type check in stopAgentInBackground.

internal/harness/tools/subagent_tools.go
```diff
@@ -0,0 +1,573 @@
...
+			Message:                 input.Task,
+			MessageActor:            actor,
+			IdempotencyKey:          "spawn:" + call.ToolCallID.String(),
+			ArchiveAfterIdleMinutes: subagent.ArchiveAfterIdleMinutes,
+			Subagent: &executionstore.SubagentLaunch{
+				ParentAgentID:       parent.ID,
+				Key:                 input.Agent,
+				MaxConcurrent:       subagent.MaxConcurrent,
+				MaxSubagents:        contract.MaxSubagents,
+				ShareParentMachines: subagent.Type == agentconfig.SubagentTypeSelf,
```
ksarangmath: Thought about subagents a bit more because something about self vs profile seemed inconsistent to me, here's a proposal (lmk what you think, and also the things below are a superset of what this current PR's logic produces, so this can all be done in a followup and doesn't require any migration in the future afaict):

I think long term, I’d want a subagent to be as close to a normal agent with minimal extra logic. Ideally 'self' vs 'profile' behave the same, 'self' would just be an easy alias for a 'profile' pointing to the same config.

Right now machine interitance differs between 'self' and 'profile' (self inherits pool machines, profile doesn't). Ideally it can be the same logic: explicitly added machines can be automatically passed down to children, but machine pool machines aren't. And the subagent tool can take in machine refs for machines that were created from its own pool (since the agent created those machines, it can also decide to give them to subagents, but explicitly granted machines, just like explicitly granted pools, are auto granted to the child). 

This design has one issue where we can't ensure that a parent's pool machine gets passed to its children. To solve this, I think we should add the ability to spin up a machine from a pool outside of an agent, and then be able to explicitly grant that pool machine to an agent, and then since it's explicitly added to that agent, its subagent will automatically inherit that machine as well. This solves the issue of the current design, where I might want a parent agent with subagents of a different profile to have the parent's pool machine, but the current design doesn't allow that. Also more concretely, for the github PR review agent I want to create, I want to have a parent agent with three subagents each with different models and system prompts (so different profiles), and all of them share a single pool machine, which can't be done with the current subagents design. 

So in the end, I think this is the only subagent-specific logic I'd want:
1. Auto-inheriting of explicitly granted machines and machine pools (but not auto-inheriting spawned pool machines, that can be an argument in the subagent tool itself)
2. Automatically sending the last message of a subagent to its parent (this PR already has that)
3. Archival of the subagent on parent archival (this PR already has that)
4. Prevention of new subagents past the max depth paramenter (this PR has this, but we could also remove the subagent tool from the child config when we create the subagent, so that the child doesn't think it should be able to create subagents. This PR also removes subagent tool from 'self', but not 'profile', which is another inconsistency)

internal/storage/executionstore/model_outputs_workflow.go
```diff
@@ -673,3 +689,10 @@ func completeSuccessfulNormalModelCallTx(
 	}
 	return nil
 }
+
+func modelOutputEndsTurn(envelope modelenvelope.ResponseEnvelope) (string, bool) {
+	if envelope.HasToolCalls() {
+		return "", false
+	}
+	return strings.TrimSpace(envelope.Text()), true
```
ksarangmath: Could we check stop_reason here as well? max_tokens returns true even though the scheduler automatically continues the child, so the parent gets an incomplete answer labeled “finished its turn.” Could we send the final-answer notification for end_turn, while still notifying the parent explicitly about refusal or content_filter? Those also come through this path. Could we cover max_tokens → end_turn in a subagent integration test?

internal/storage/executionstore/subagents_store.go
```diff
@@ -0,0 +1,896 @@
...
+	default:
+		return SubagentStatus{}, storeerr.InvalidRequest(
+			fmt.Errorf("%d subagents match %q; address it by full agent id instead", len(matches), reference),
+		)
+	}
+}
+
+// SubagentRef is the short form of an agent id that subagent tools accept:
+// the trailing characters of the public id, which carry its random bits.
+func SubagentRef(agentID ID) string {
```
ksarangmath: Could we change this logic to using the raw public agent id? I'm planning to change machine ref to do the same. I looked at codex and opencode, and they use full UUIDs to refer to subagents, and also saw some papers where even crappy LLMs like llama get nearly 100% accuracy in retrieving UUIDs, so I think it'd be better to just use the same public id that users also can use.

internal/harness/tools/subagent_tools.go
```diff
@@ -0,0 +1,573 @@
...
+	}
+	summary, err := subagentSummaryFromStatus(status)
+	if err != nil {
+		return nil, err
+	}
+	entries := make([]subagentEventSummary, 0, len(events))
+	for _, event := range events {
+		entries = append(entries, subagentEventSummaryFromRecord(event))
+	}
+	content, err := structuredToolResultContent(map[string]any{
```
ksarangmath, line 404: Could read_agent return the same responses as our turns API? By default it could return a small page of turns, and when given a turn_id, return a page of events from that turn, with the same IDs and pagination as the API.

That lets the model get an overview first and go deeper when needed. Could we expose the needed reads through ToolCallReader and share the response-building code with the API, so we don’t maintain a separate history format just for subagents?

internal/toolcatalog/subagent_tools.go
```diff
@@ -0,0 +1,123 @@
...
+		"background and its final answer arrives later as a message from it; read_agent shows its progress. Never " +
+		"guess or predict a pending subagent's result. Its answer is not shown to the user, so relay what matters. " +
+		"`name` is a display label only; address the subagent by the `agent_ref` in the result."
+	readAgentToolDescription = "Read a subagent's timeline by agent_ref: its inputs, model outputs, and tool results " +
+		"with text content. Use it to check on progress or to fetch a finished subagent's answer without waiting " +
+		"for its message. Page forward with `after_sequence` (oldest first) or backward from the end with " +
+		"`before_sequence` (newest first). Never guess at a subagent's result; read it or wait for its message."
+	sendAgentMessageToolDescription = "Send a message to one of your subagents by agent_ref. Your plain text " +
+		"output is not visible to subagents; this tool is the only way to reach them. The message interrupts " +
+		"whatever the subagent is doing and cancels any question or permission request it has open; it cannot " +
```
ksarangmath: Could we clarify that the child picks up this message after its current model call and tool batch finish, it doesn't interrupt that? “Interrupts whatever the subagent is doing” sounds like it stops running work, but this only cancels open questions and permission requests. The same wording appears in the sendAgentMessage tool result.

internal/toolcatalog/subagent_tools.go
```diff
@@ -0,0 +1,123 @@
...
+			"message": map[string]any{
+				"type":        "string",
+				"minLength":   1,
+				"description": "Message text delivered to the subagent.",
+			},
+		},
+	)
+}
+
+func stopAgentTool() (Entry, error) {
```
ksarangmath, line 115: Could stop_agent take an archive flag that defaults to false? By default, it could cancel the target agent’s current work and leave it available to message again. With archive: true, it should keep the current behavior of archiving the agent and its descendants.

internal/storage/executionstore/model_outputs_terminal_error.go
ksarangmath: Could we share the transaction setup that takes the parent/child locks before validating runtime ownership, and use it for the remaining failure paths too? Compaction failures and admission failures during initial calls, retries, and compaction handoffs still lock the child first and then reach this parent notification, so they can deadlock with archival.

internal/storage/queries/subagents.sql
```diff
@@ -0,0 +1,242 @@
...
+        OR EXISTS (
+          SELECT 1
+          FROM agent_wakeups wake
+          WHERE wake.agent_id = subtree.id
+        )
+        OR EXISTS (
+          SELECT 1
+          FROM tool_calls call
+          WHERE call.agent_id = subtree.id
+            AND call.state IN ('running', 'waiting', 'awaiting_permission')
```
ksarangmath: Could both idle-archive queries use agent_has_incomplete_tool_batch instead of maintaining their own list of unfinished tool states? A custom tool waiting for an external result stays ready, so these checks miss it and can archive the child before its result arrives.

internal/storage/executionstore/subagents_store.go
```diff
@@ -0,0 +1,896 @@
...
+) ToolCallCommand {
+	return toolCallCommandFunc(func(ctx context.Context, tx *toolCallTransaction) (any, error) {
+		child, err := loadAgentInProjectTx(ctx, tx.tx, tx.input.ProjectID, targetAgentID)
+		if err != nil {
+			return nil, err
+		}
+		if child.ParentAgentID != tx.input.AgentID {
+			return nil, storeerr.InvalidRequest(errors.New("target agent is not a subagent of this agent"))
+		}
+		machines, err := archiveAgentTreeTx(ctx, tx.tx, tx.q, tx.notifications, child.ProjectID, child.ID, nil, "")
```
ksarangmath: nit: Could stop_agent and idle archival acquire the organization/project lifecycle gates before locking the tree, using shared setup with ArchiveAgent? These paths currently skip those gates, so they can deadlock with project or organization deletion.

api/openapi/openapi.yaml
```diff
@@ -8851,12 +8869,34 @@ components:
...
+      properties:
+        id:
+          $ref: '#/components/schemas/AgentID'
+        name:
+          $ref: '#/components/schemas/AgentName'
+        key:
+          type: string
+        state:
+          type: string
+          enum: [running, idle, waiting_on_human, archived]
```
ksarangmath: Could we call this waiting_on_interaction instead? I don't think there should be concept of "human"

internal/httpapi/agent_routes.go
```diff
@@ -653,6 +643,16 @@ func (s strictOpenAPIServer) listAgents(
...
+		parentAgentID, err := publicid.Decode(publicid.KindAgent, *params.ParentAgentId)
+		if err != nil {
+			return nil, apierror.FromCode(openapi.ErrorCodeInvalidRequest, "invalid parent agent filter")
+		}
+		filters.ParentAgentID = &parentAgentID
+	}
+	if params.IncludeSubagents != nil {
+		filters.IncludeSubagents = *params.IncludeSubagents
+	}
 	list, err := parseResourceListQuery(resourceListQueryInput{
```
ksarangmath, line 665: nit: Could we include parent_agent_id, include_subagents, and agent_profile_id when creating and checking the cursor, using the pagination helper’s Extra field? Currently, a cursor from one parent is accepted for another parent and can silently skip some of its children.

internal/httpapi/agent_routes.go
```diff
@@ -990,6 +981,24 @@ func (s *Server) currentAgentResponse(
 	if err != nil {
 		return openapi.GetAgentResponse{}, apierror.ProjectScoped(err)
 	}
+	subagents, err := s.store.Execution().ListSubagents(ctx, record.ProjectID, record.ID)
```
ksarangmath: Could we load children through the paginated listAgents(parent_agent_id=...) endpoint instead of including every child in GetAgent? We could add the status fields that the UI needs and an option to include archived children.

internal/httpapi/agent_interaction_routes.go
```diff
@@ -39,11 +39,19 @@ func (s strictOpenAPIServer) listAgentInteractions(
 	if params.State != nil {
 		state = executionstore.AgentInteractionState(*params.State)
 	}
-	page, err := s.server.store.Execution().ListAgentInteractionsForAgent(
+	agentIDs := []storage.ID{agentID}
+	if params.IncludeSubagents != nil && *params.IncludeSubagents {
```
ksarangmath, line 43: nit: Could we add a test that pages through interactions from a parent, child, and grandchild with include_subagents=true and makes sure everything is correct

internal/storage/queries/subagents.sql
```diff
@@ -0,0 +1,242 @@
...
+           SELECT 1
+           FROM agent_runtime_locks runtime_lock
+           WHERE runtime_lock.agent_id = agent.id
+         )
+         OR EXISTS (
+           SELECT 1
+           FROM agent_wakeups wake
+           WHERE wake.agent_id = agent.id
+         )
+         OR agent_next_wakeup_ready_at(agent.project_id, agent.id) IS NOT NULL
```
ksarangmath, line 38: Could ListChildAgents also account for unfinished tool calls using agent_has_incomplete_tool_batch? A child waiting for an external custom-tool result can currently appear idle. This is the same case as the idle-archival comment, but it also affects the status returned to the parent and the UI.

api/openapi/openapi.yaml
```diff
@@ -3946,6 +3958,12 @@ paths:
           required: false
           schema:
             $ref: '#/components/schemas/AgentInteractionState'
+        - name: include_subagents
```
ksarangmath, line 3975: Right now I don't think there's a great way to detect whether subagents have called a custom tool or an interaction. We could add subagent custom tools and interactions to the root SSE stream, and have the agent id attached for people to know if it was a subagent or not? I don't think we should put regular tool calls/reasoning/text in the sse though for a parent. Outbound webhook would be the cleaner longer term thing to do.
Also the GET for tool calls should probably have the option to list all subagent tool calls as well

frontend/packages/react/src/domains/agent-interactions.ts
```diff
@@ -10,7 +10,7 @@ import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query'
 import { useOmnaraClient } from '../omnara-client'
 import { agentInputBacklogQueryKey } from './agent-input-backlog'
 
-const openInteractionsQuery = { state: 'open', limit: 100 } as const
+const openInteractionsQuery = { state: 'open', limit: 100, include_subagents: true } as const
```
ksarangmath, line 13: Right now, the page checks for questions and permission requests every second while the parent is working, then stops. But a child can keep working after the parent finishes and ask for permission later, so that request can stay hidden until someone refreshes the page.

We could refresh the displayed requests when a SSE notification says a question or permission request was created, answered, or canceled. And if something is missed over sse, a refresh of the page can call the static endpoints

frontend/apps/web/src/routes/AgentView.tsx
ksarangmath: might need to update this based on also getting subagent interactions now

internal/agentconfig/subagents.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+	return compiled, nil
+}
+
+func validateSubagentToolConfiguration(source AgentConfigSource) error {
+	for name := range source.Tools {
+		if !toolcatalog.IsSubagentToolName(name) {
+			continue
+		}
+		if len(source.Subagents) == 0 {
```
ksarangmath: nit: could this just apply to spawn agent? I think someone should be able to remove subagents from config but still access previous subagents

ksarangmath (approved):
Turned out to be more comments than I thought as I kept thinking about subagents, but I think most are straightforward and I mention in the larger comment that the current design is a subset of what a broader design can look like, so we can merge this in and then have followups for the rest if you think those make sense.

internal/harness/tools/subagent_tools.go
cmsparks: Going to do 2 things here:
1. Add max depth parameter
2. Default to inherit for both profile and self (including machine pool machines. Later, we'll do 2 things: a) implicitly set machine related tools b) allow not inheriting machines for profiles (a is probably a soft prereq to b?)

We'll keep the instruction + model overrides

internal/harness/tools/subagent_tools.go
cmsparks: yeah we definitely should do that, good point

internal/agentconfigcompile/compile.go
cmsparks: Just going to regenerate the yaml. this loses comments + whitespace, but I think it should be fine

internal/storage/executionstore/subagents_store.go
cmsparks: sg, yeah as long as we're consistent in the codebase i'll remove

api/openapi/openapi.yaml
cmsparks: Yeah was considering doing that, I'll add a parameter that lets us stream back the interactions and custom tools back

internal/httpapi/agent_event_stream_subagents.go
```diff
@@ -0,0 +1,80 @@
...
+				case s.updates <- update:
+				default:
+					if s.server.log != nil {
+						s.server.log.Debug(
+							"drop subagent tool call update because subscriber buffer is full",
+							"agent_id",
+							agentID,
+						)
+					}
+				}
```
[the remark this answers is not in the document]
cmsparks: dropping when full is fine, best effort

PR #477, opened 2026-09-03
docs: Add reddit-signal-agent and linkedin-signal-agent examples

ishaansehgal99:
## What

Two new build-your-own listening-agent notebooks in `examples/`, following the `x-signal-agent` pattern, plus docs to surface them.

### `reddit-signal-agent`
A daily Reddit digest with **no Reddit API key** (Reddit gates its Data API behind manual approval). The agent fetches through the hosted Apify MCP server (`trudax/reddit-scraper-lite`) wired into the agent config's `mcp` block with a bearer project secret — the MCP server is the entire fetch layer, so no machine pool is needed. It sweeps keyword searches plus the new-post feeds of high-signal subreddits, filters hard (same builder-signal taxonomy as the X agent), and delivers a ≤5-item digest to Slack or the console. Capped at 120 results ≈ $0.40/scan against Apify's free $5/month.

### `linkedin-signal-agent`
The same daily digest for LinkedIn — **no LinkedIn API, account, or session cookie**. Two no-cookie Apify scrapers (`harvestapi/linkedin-post-search` for keyword search, `harvestapi/linkedin-profile-posts` for tracked company/profile feeds) behind one MCP server URL.

### Docs
Both examples added to the Mintlify examples overview table alongside the X agent.

### Housekeeping
`@omnara/sdk` pinned at `^0.1.7` in the signal-agent examples: 0.1.8 published <24h ago and is blocked by Deno's default `minimumDependencyAge` supply-chain gate; the caret range picks it up automatically once it ages past the window.

## Notes
- Notebooks are idempotent — rerun any cell after editing and resources update in place.
- Secrets never land in the notebooks: tokens live in a gitignored `.env` and are upserted as Omnara project secrets; only resource IDs appear in outputs.
- Both examples verified end-to-end against live Apify runs and deployed profiles.


examples/linkedin-signal-agent/deno.json
```diff
@@ -0,0 +1,7 @@
+{
```
cmsparks: can we swap to just normal package.json which deno should be compatible with (just to be consistent with the rest of the repo)

examples/linkedin-signal-agent/deno.json
```diff
@@ -0,0 +1,7 @@
...
+  "imports": {
+    "@omnara/sdk": "npm:@omnara/sdk@^0.1.7",
```
cmsparks: we should probably reference the omnara SDK within the workspace, rather than the published package. 

something like:
```
{
  "private": true,
  "type": "module",
  "dependencies": {
    "@omnara/sdk": "file:../../frontend/packages/sdk",
    "dotenv": "^16.4.0"
  }
}
```


that helps ensure that our demos are type safe

examples/linkedin-signal-agent/demo.ipynb
```diff
@@ -0,0 +1,710 @@
+{
```
cmsparks: suggestion, not required, but it might be nicer for dev X and typechecking if we used the jupytext notebook format instead: https://jupytext.org/ and https://github.com/jupytext/jupytext

Then, demo just becomes a normal typescript file (with annotations), which the user can directly run, or alternatively could be viewed in jupyter. This has 2 benefits:
* no need to install jupyter + deno if a user doesn't want to, they can just use pnpm or something
* we get CI checks on the typescript so we can verify that our SDK updates don't break demos

cmsparks (approved):
approved

examples/linkedin-signal-agent/demo.ipynb
ishaansehgal99: nice this is exactly what i was looking for - agree

examples/linkedin-signal-agent/deno.json
ishaansehgal99: i think these examples should stay on the published @omnara/sdk. they're meant to be user copies. someone follows the docs, or points their agent at them. from sitting with users, agents (especially the weaker models people run) do some very dumb things with examples that don't work standalone. path that only resolves inside our monorepo is the kind of thing they'll trip on. pointing at upstream can also conflict with what's actually released and what users are running: the workspace package is unreleased head, so the examples could compile against APIs that aren't on npm yet. it also doesn't change CI today nothing installs or type-checks examples/. in a future PR it'd make sense to add a CI check, but i think that check should typecheck against the latest published version: if it doesn't, the examples can reflect something newer than what anyone actually has. i also checked what other projects do, next.js ("next": "latest"), supabase, and tanstack query all pin their examples to published versions. the projects that use local refs (openai-node, langchainjs, vercel/ai) treat their examples as internal test harnesses wired into repo CI, which isn't what these are

examples/linkedin-signal-agent/demo.ts
```diff
@@ -0,0 +1,498 @@
...
+text for a human to post — helpful, never salesy, always disclosing the
+affiliation. Never post to LinkedIn yourself.
+`,
+  model: {
+    provider_config: 'omnara-openrouter', // default model provider config in your org
+    name: 'openai/gpt-5.6-sol', // configured model name on that provider config
+  },
+  mcp: {
+    linkedin: {
+      url: 'https://mcp.apify.com/?tools=harvestapi/linkedin-post-search,harvestapi/linkedin-profile-posts',
```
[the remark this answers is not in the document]
ishaansehgal99: false positive — the Apify MCP server auto-injects the run helpers when an Actor tool is loaded. from the [apify-mcp-server README](https://github.com/apify/apify-mcp-server): "When `call-actor`, an Actor tool, or `get-actor-run` is present, the server auto-injects `get-actor-run`, `get-dataset-items`, `get-key-value-store-record`, and `abort-actor-run`." verified live: both the reddit and linkedin agents ran end-to-end with these exact URLs, polling `get-actor-run` and fetching with `get-dataset-items` successfully.

examples/linkedin-signal-agent/demo.ts
```diff
@@ -0,0 +1,502 @@
...
+// %%
+import { load } from 'jsr:@std/dotenv'
+import { bearerToken, createOmnaraClient, openAgentEventStream, sdk } from '@omnara/sdk'
+
+const env = await load()
+
+const client = createOmnaraClient({
+  baseUrl: 'https://api.omnara.com/v1',
+  auth: bearerToken(env.OMNARA_API_KEY),
+})
```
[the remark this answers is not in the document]
ishaansehgal99: false positive — this checks the workspace SDK's generated paths (`/api/v1/...`) but the examples resolve the published `@omnara/sdk@0.1.8` from npm, whose dist dropped the prefix (paths like `/me`) and moved it into the base URL; 0.1.8's README documents `baseUrl: 'https://api.omnara.com/v1'`. verified live today: full end-to-end runs of the X and LinkedIn demos plus auth smoke tests from all four folders on deno/node/bun, all against this base URL. (the old `https://app.omnara.com` base is what actually broke once `^0.1.7` started resolving 0.1.8 — this change is the fix.)

PR #478, opened 2026-09-03
fix: center last-used badge and stabilize a flaky integration test

asherdale:
- vertically center the `Last used` badge within the password sign-in button while preserving the OAuth provider badge layout
- replace cross-clock timing bounds with explicit before-and-after tool-call state assertions for daemon process-start coverage


ksarangmath (approved):
LGTM

PR #479, opened 2026-09-03
fix: explain MCP tool names that exceed the model limit

cmsparks:
`toolcatalog.ValidateMCPRuntimeToolName` returns a typed `MCPRuntimeToolNameTooLongError` whose message shows the prefixed name, its length, the 64-character limit, and the maximum server name length that would make it fit (or that the tool name is too long under any server name).
`validateDiscoveredTools` wraps that validation error, so the connection's `initialize_error` explains the length problem instead of the generic message.

Frontend: mcp connection panel shows an error for tools whose prefixed name is too long
Frontend: Failed MCP connections in the agent sidebar show a `Failed` badge with the error inline

### DIFF
--- a/frontend/apps/web/src/components/agents/AgentConfigMcpServerTools.tsx
+++ b/frontend/apps/web/src/components/agents/AgentConfigMcpServerTools.tsx
@@ -10,10 +10,13 @@ import {
 import { useState } from 'react'
 
 import { mcpServerToolsRequest } from '@/components/agents/mcpServerToolsRequest'
-import type {
-  BasicMcpServer,
-  BasicMcpTool,
-  McpAuthType,
+import {
+  type BasicMcpServer,
+  type BasicMcpTool,
+  type McpAuthType,
+  mcpRuntimeToolName,
+  mcpRuntimeToolNameError,
+  mcpRuntimeToolNameMaxLength,
 } from '@/components/agents/useAgentBuilderForm'
 import { SearchIcon, Trash2Icon, TriangleAlert } from '@/components/icons'
 import { Button } from '@/components/ui/button'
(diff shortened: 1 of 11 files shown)


frontend/apps/web/src/components/agents/AgentConfigMcpServerTools.tsx
```diff
@@ -54,40 +50,25 @@ export function AgentConfigMcpServerTools({
...
-
-  function updateOverride(name: string, patch: Partial<Omit<BasicMcpTool, 'name'>>) {
-    onToolsChange(server.tools.map((tool) => (tool.name === name ? { ...tool, ...patch } : tool)))
-  }
-
+  const unexposableTools = discovered.flatMap((tool) => {
+    if (!mcpToolEnabled(server, tool.name)) return []
+    const error = mcpRuntimeToolNameError(server.name, tool.name)
+    return error === undefined ? [] : [{ name: tool.name, error }]
+  })
```
ishaansehgal99: nit: I think this only catches the length case. do we want to also handle other invalid tools like dots, spaces etc. which may fail the connection at runtime but shows no warning here. Fine as a follow-up.

frontend/apps/web/src/components/agents/AgentConfigMcpServerTools.tsx
```diff
@@ -54,40 +50,25 @@ export function AgentConfigMcpServerTools({
...
-    if (overriddenNames.has(name)) return
-    onToolsChange([...server.tools, { name, enabled: null, permission: null }])
-    setQuery('')
-  }
-
-  function updateOverride(name: string, patch: Partial<Omit<BasicMcpTool, 'name'>>) {
-    onToolsChange(server.tools.map((tool) => (tool.name === name ? { ...tool, ...patch } : tool)))
-  }
-
+  const unexposableTools = discovered.flatMap((tool) => {
```
ishaansehgal99: nit: this only checks discovered tools. a manually entered override can also be enabled with a name over the 64-character limit, but it won’t appear here, so the user only finds out when saving fails. should we check configured overrides too, while still allowing them when disabled?

ishaansehgal99 (approved):

ishaansehgal99 (approved):
two very minor optional nits

PR #486, opened 2026-09-04
feat: add SigV4 authentication for Bedrock providers

asherdale:
- add SigV4 authentication for Amazon Bedrock model providers while preserving API-key authentication
- persist the AWS signing service and region with each SigV4 provider configuration
- reuse the shared AWS signer across model providers and MCP connections, including session credentials and optional role assumption
- add web flows for creating and selecting AWS credential secrets when configuring Bedrock providers
- expose the new authentication fields through the OpenAPI contract and generated SDK


cmsparks (approved):
meant to approve earlier, but lgtm

PR #495, opened 2026-09-09
feat: add path-based file transfer tools

asherdale:
- Replace `upload_artifact` and `download_artifact` with `upload_file` and `download_file`, using `/artifacts` for uploads and `/artifacts/<artifact_id>` for downloads.
- Keep the existing machine-transfer lifecycle, approvals, timeouts, and background wake behavior; call the shared transfer implementation directly.
- Migrate existing configs in place, preserving IDs, tool settings, and permissions while rewriting source and definitions and recomputing hashes.
- Bind daemon callbacks to the exact tool and recorded resource path, and return the created path after an upload.
- Update templates, tool selection, launch guidance, approval prompts, and documentation to use the new tools.
- Limit this change to artifact transfers; skills and other resource types remain out of scope.


internal/toolcatalog/catalog.go
```diff
@@ -261,6 +267,12 @@ func buildDefaultCatalog() (Catalog, error) {
 	if entries[ToolNameWebFetch], err = webFetchTool(); err != nil {
 		return Catalog{}, err
 	}
+	if entries[ToolNameUploadFile], err = uploadFileTool(machineRef); err != nil {
+		return Catalog{}, err
+	}
+	if entries[ToolNameDownloadFile], err = downloadFileTool(machineRef); err != nil {
+		return Catalog{}, err
+	}
 	if entries[ToolNameUploadArtifact], err = uploadArtifactTool(machineRef); err != nil {
```
ksarangmath: Should we do a hard cutover in this PR? We can migrate existing configs to the new tools with their permissions intact, and stop publishing and accepting the old tools in the backend too. We can adhoc check if any agents were created in the deploy window with the old tools and migrate them.

internal/harness/tools/vfs_tools.go
```diff
@@ -0,0 +1,214 @@
...
+	machineRef, err := resolveOptionalMachineRef(input.MachineRef)
+	if err != nil {
+		return resolvedDownloadFileRequest{}, err
+	}
+	return resolvedDownloadFileRequest{
+		ArtifactID: artifactID, Destination: input.Destination, MachineRef: machineRef,
+	}, nil
+}
+
+func artifactUploadCall(call model.ToolCall, resolved resolvedUploadFileRequest) (model.ToolCall, error) {
```
ksarangmath: Could the new tools call the shared transfer implementation directly? They currently construct the old tools’ JSON inputs, which the old handlers then parse again. That ties the new interface to the interface we’re removing.

internal/toolcatalog/catalog.go
```diff
@@ -493,6 +505,28 @@ func uploadArtifactTool(machineRef map[string]any) (Entry, error) {
 	)
 }
 
+func uploadFileTool(machineRef map[string]any) (Entry, error) {
+	return toolEntry(
+		ToolNameUploadFile,
+		uploadFileToolDescription,
+		[]string{"path", "source"},
+		map[string]any{
+			"path": map[string]any{
```
ksarangmath, line 502: nit: is there a way to enforce a prefix here? to be artifacts/?

internal/toolcatalog/catalog.go
```diff
@@ -57,6 +57,12 @@ const (
 		"The result includes artifact metadata after a successful upload."
 	downloadArtifactToolDescription = "Copy an existing artifact to an attached machine. " +
 		"The result includes a process_id; use the process tools to inspect the transfer if it is still running."
+	uploadFileToolDescription = "Copy a file from an attached machine into Omnara's virtual filesystem. " +
+		"Currently supports creating artifacts at /artifacts. The file must be regular, non-empty, and at most 10 MiB. " +
+		"Successful uploads return the artifact_id."
```
ksarangmath: should we return the path instead of artifact id?

internal/storage/queries/processes.sql
```diff
@@ -394,7 +394,8 @@ WHERE process.org_id = sqlc.arg(org_id)
 -- name: GetDaemonArtifactProcessScope :one
 SELECT process.project_id,
        process.agent_id,
-       COALESCE(tool_call.input->>'artifact_id', '')::text AS artifact_id
+       tool_call.name AS tool_name,
+       COALESCE(tool_call.input->>'artifact_id', tool_call.input->>'path', '')::text AS artifact_locator
```
ksarangmath: If we do a hard cutover to the new tool schema, then we can remove this complexity and artifact_locator can just have a single meaning.

internal/harness/tools/vfs_tools_test.go
```diff
@@ -0,0 +1,139 @@
...
+		t.Run(test.name, func(t *testing.T) {
+			_, err := resolveDownloadFileRequest(json.RawMessage(test.raw))
+			if err == nil || !strings.Contains(err.Error(), test.want) {
+				t.Fatalf("error = %v, want %q", err, test.want)
+			}
+		})
+	}
+}
+
+func TestVFSArtifactCallsReuseLegacyInputs(t *testing.T) {
```
ksarangmath: nit: Could we add coverage that takes both new tools through approval and executor dispatch? The current tests check input conversion or create processes directly, so they don’t verify that the approved operation becomes the expected transfer.

internal/storage/executionstore/tool_calls_workflow.go
```diff
@@ -235,8 +235,19 @@ func commandTerminalToolResult(
...
+	if call.Name == toolcatalog.ToolNameUploadArtifact {
+		return true
+	}
+	if call.Name != toolcatalog.ToolNameUploadFile {
+		return false
+	}
+	var input struct {
+		Path string `json:"path"`
+	}
+	return json.Unmarshal(call.Input, &input) == nil && input.Path == toolcatalog.ArtifactVFSRoot
```
ksarangmath, line 247: Can we make the transfer’s purpose and its relationship to the artifact explicit here? Completion currently infers them from the tool name/input and an idempotency-key naming convention. Since we’re changing this interface and doing a hard cutover, this seems like a good place to clean up that model too.

ksarangmath (commented):
The new tools look good to me, only thing is I think we can probably do a hard cutover to the new tools and migrate old agents to it. Is there a reason to keep both right now?

internal/toolcatalog/catalog.go
asherdale: done - removed the old tools and added an in-place config migration preserving settings and permissions

internal/harness/tools/vfs_tools.go
asherdale: good call, both tools call the shared transfer implementation directly now. removed the legacy request types and JSON round trip

internal/toolcatalog/catalog.go
asherdale: also good call, upload now uses an enum for `/artifacts`, and download uses a pattern for `/artifacts/<artifact_id>`. both are also validated server-side

internal/toolcatalog/catalog.go
asherdale: updated this, the upload result now returns `path` instead of `artifact_id`

internal/storage/queries/processes.sql
asherdale: renamed it to `path` and removed the legacy branches

internal/storage/executionstore/tool_calls_workflow.go
asherdale: as discussed, leaving this unchanged for now. in the future we can centralize the tool name + path check, especially as we add more path possibilities in the future

internal/toolcatalog/catalog.go
```diff
@@ -508,21 +516,23 @@ func uploadArtifactTool(machineRef map[string]any) (Entry, error) {
...
+		downloadFileToolDescription,
+		[]string{"path", "destination"},
 		map[string]any{
-			"artifact_id": map[string]any{
+			"path": map[string]any{
 				"type":        "string",
 				"minLength":   1,
-				"description": "Public artifact_id provided alongside a conversation attachment.",
+				"description": "Source path in Omnara's virtual filesystem. Currently supports /artifacts/<artifact_id>.",
+				"pattern":     "^" + ArtifactVFSRoot + "/art_[a-zA-Z2-7]{26}$",
```
ksarangmath, line 529: nit: I changed the public ids to accept only lowercase in a PR just now, so this can be changed to use lowercase only

internal/toolcatalog/catalog.go
ksarangmath, line 380: nit: don't have to change in this PR, probably can change later in my integration pr or your memory pr, but should this take a path?

ksarangmath (approved):
LGTM!

internal/toolcatalog/catalog.go
asherdale: will handle this in the upcoming memory PR

PR #496, opened 2026-09-09
feat: Implement stateless MCP client with shared server data cache

cmsparks:


migrations/000032_mcp_catalog_refresh_errors.sql
```diff
@@ -0,0 +1,4 @@
+-- +goose Up
+
+ALTER TABLE mcp_server_catalogs
+    ADD COLUMN refresh_error text NOT NULL DEFAULT '';
```
asherdale: minor nit: can we fold this into `migrations/000031_mcp_server_catalogs.sql` instead of a second migration? easier to see the initial table shape at once

internal/mcp/catalog.go
```diff
@@ -0,0 +1,624 @@
...
+		if err != nil {
+			return 0, fmt.Errorf("allocate mcp request sequence: %w", err)
+		}
+		return seq, nil
+	}
+}
+
+func (m Manager) legacyCatalogFetch(session legacySession) catalogFetch {
+	return func(ctx context.Context, current executionstore.MCPServerCatalogRecord) (catalogContents, error) {
+		listing, err := listAllTools(ctx, m.Client, session.Conn, session.NextRequestID)
```
asherdale, line 158: if the server dropped this session, `ErrSessionExpired` isn't handled here and the turn fails. should we reopen a session and retry instead (like `ephemeralLegacyCatalogFetch`)?

internal/mcp/catalog.go
```diff
@@ -0,0 +1,624 @@
...
+	if err := callerCtx.Err(); err != nil {
+		return executionstore.MCPServerCatalogRecord{}, err
+	}
+	ctx, cancel := context.WithTimeout(callerCtx, timeout)
+	defer cancel()
+	contents, err := fetch(ctx, current)
+	if err != nil {
+		return executionstore.MCPServerCatalogRecord{}, errors.Join(
+			err,
+			m.markCatalogRefreshFailed(callerCtx, identity, current, owner, err),
```
asherdale: caller cancellation gets persisted as `refresh_error` and fails every reader in the org, should we run this on `WithoutCancel` like the OAuth lease owner does to avoid the same thing?

internal/mcp/connection_manager.go
```diff
@@ -254,63 +315,29 @@ func (m Manager) initialize(
...
 		)
 	}
-	tools, err := listAllTools(ctx, m.Client, wireConn, func(ctx context.Context) (int64, error) {
-		seq, err := m.Execution.NextMCPRequestSequence(ctx, projectID, agentID, conn.ID)
-		if err != nil {
-			return 0, fmt.Errorf("allocate mcp tools/list request sequence for %q: %w", conn.ServerKey, err)
-		}
-		return seq, nil
-	})
+	probe, err := m.probeServer(ctx, wireConn)
```
asherdale: minor nit: if the catalog is already fetched, fresh, and legacy, we could skip the discover probe here and go straight to `initializeLegacy` to save a rejected request on every legacy session reinit?

internal/mcp/catalog.go
asherdale, line 417: caller cancellation gets persisted as `refresh_error` and fails every connection sharing this catalog, should we run this on `WithoutCancel` like the OAuth lease owner does to avoid the same thing?

asherdale (approved):
lgtm 👌

migrations/000032_mcp_catalog_refresh_errors.sql
cmsparks: you're right, adding this

internal/mcp/catalog.go
```diff
@@ -0,0 +1,630 @@
...
+					current.ID,
+					owner,
+				)
+				return current, nil
+			}
+			ownerTimeout := leaseTTL - time.Since(leaseAttemptStarted) - catalogRefreshOwnerHeadroom
+			return m.fetchCatalogAsLeaseOwner(ctx, identity, current, owner, ownerTimeout, fetch)
+		}
+		if current.RefreshError != "" {
+			return executionstore.MCPServerCatalogRecord{}, errors.New(current.RefreshError)
```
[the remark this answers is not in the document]
cmsparks: i think this is probably fine?

PR #499, opened 2026-09-10
feat: add generated mcp implementation

cmsparks:


internal/httpapi/apimcp/server.go
```diff
@@ -0,0 +1,603 @@
...
+	if options.Grants != nil {
+		server.AddReceivingMiddleware(grantMiddleware(options.Grants, operationByTool))
+	}
+	return server, nil
+}
+
+func NewHandler(server *mcp.Server) http.Handler {
+	return mcp.NewStreamableHTTPHandler(
+		func(*http.Request) *mcp.Server { return server },
+		&mcp.StreamableHTTPOptions{
```
asherdale, line 136: can we bump both body limits for attachments? the outer 1 MiB cap rejects files the REST endpoint accepts, and the SDK's 4 MiB default would still block larger ones

internal/httpapi/apimcp/manifest.go
```diff
@@ -0,0 +1,96 @@
...
+	{Name: "orgs_list", OperationID: "listOrganizations"},
+	{Name: "orgs_create", OperationID: "createOrganization"},
+	{Name: "orgs_overview", OperationID: "getOrgOverview"},
+
+	{Name: "projects_list", OperationID: "listVisibleProjects"},
+	{Name: "projects_create", OperationID: "createProject"},
+
+	{Name: "agents_list", OperationID: "listAgents"},
+	{Name: "agents_get", OperationID: "getAgent"},
+	{Name: "agents_launch", OperationID: "createAgent"},
```
asherdale, line 22: can we add a tool mapping to `createAgentConfig`? launch and profile creation need a config ID, but there's no MCP tool to make one, so you can't get started in a fresh project

asherdale (approved):
LGTM

internal/httpapi/apimcp/server.go
cmsparks: I deliberately kept it at 1MB for now because generally I wouldn't expect the model to be directly outputting 1mb (and it'd probably be an antipattern if so)

PR #503, opened 2026-09-10
feat: add Modal sandbox support

asherdale:
- add Modal as a bring-your-own sandbox provider using one secret containing token_id and token_secret
- support retry-safe provisioning, ownership-checked deletion, and optional runtime protection through the existing machine-pool lifecycle
- default the optional Modal app to omnara and create it on first provisioning, with configurable images, regions, and resource limits
- expose Modal pool configuration in the dashboard and API, and document setup and the 24-hour sandbox lifetime


internal/machinepool/providers/modal/provider.go
```diff
@@ -0,0 +1,229 @@
...
+	if providerResourceID != "" {
+		target, found, err = api.GetSandboxByID(ctx, providerResourceID)
+	} else {
+		target, found, err = api.GetSandboxByName(ctx, expectedName)
+	}
+	if err != nil || !found {
+		return "", false, err
+	}
+	if !target.Running {
+		return "", false, nil
```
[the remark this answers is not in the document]
asherdale: not applicable to Modal’s lifecycle contract: name lookup only returns running sandboxes, and names become reusable once the sandbox stops, including after timeout. no additional deletion is needed

internal/machinepool/providers/modal/client.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+		return sandbox{}, false, err
+	}
+	return describeSandbox(ctx, target)
+}
+
+func (c *modalAPI) DeleteSandbox(ctx context.Context, id string) error {
+	target, err := c.client.Sandboxes.FromID(ctx, id, nil)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
asherdale: not applicable to the pinned SDK: `FromID` only constructs a local handle and always returns a nil error. the actual RPCs (`GetTags`, `Poll`, and `Terminate`) already normalize not-found, with test coverage

internal/machinepool/providers/modal/client.go
```diff
@@ -0,0 +1,171 @@
+package modal
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	modalsdk "github.com/modal-labs/modal-client/go"
```
cmsparks: This is a pretty thin wrapper over the modal SDK, I'd suggest we either just do raw HTTP requests like we do for the other providers or use the SDK client directly in provider.go (and then we can remove client.go and client_test.go)

cmsparks: Ah I see it's actually a grpc API, I think we can probably remove client.go, and then just call the client functions directly in provider.go instead of wrapping them?

cmsparks (approved):

internal/machinepool/providers/modal/client.go
asherdale: good call, removed `client.go` and now call the client functions directly in `provider.go`

PR #504, opened 2026-09-10
fix(model): explicitly disable strict function tool generation

ksarangmath:
Preserve optional properties and general JSON Schemas in function-tool declarations by explicitly sending `strict: false` for Responses requests and OpenRouter Chat Completions. Responses can otherwise normalize schemas into strict mode. Other Chat Completions variants retain their existing omission behavior.

Use the existing Chat compatibility policy to select field emission. Separate function definitions from function calls so declaration-only fields stay out of conversation replay.

Validation:

- Shared request tests cover explicit false versus omission across all five supported format/variant combinations, schema preservation, and parameterless tools. Existing history and replay tests cover declaration-field isolation.
- Model and model-context unit tests, race tests, vet, formatting, and targeted repository lint.
- The existing live OpenRouter tool-call smoke test confirms request acceptance and a returned tool call.

### DIFF
--- a/internal/model/function_tools_wire_test.go
+++ b/internal/model/function_tools_wire_test.go
@@ -1,0 +1,69 @@
+package model_test
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/omnara-ai/omnara/internal/model"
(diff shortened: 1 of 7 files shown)


cmsparks (approved):
lgtm

PR #505, opened 2026-09-10
feat: update secret values from the dashboard

asherdale:
- Allow renaming secrets and replacing generic, AWS, and OAuth credentials from one Edit dialog using the existing APIs.
- Keep stored values hidden and require complete credential replacements, with Clear and Undo controls for field changes.
- Preserve replacement drafts across list refreshes and partial failures so retries do not repeat a successful rename.

### DIFF
--- a/frontend/apps/web/src/components/org/EditSecretDialog.tsx
+++ b/frontend/apps/web/src/components/org/EditSecretDialog.tsx
@@ -1,4 +1,4 @@
-import { useUpdateSecret } from '@omnara/react'
+import { useCreateSecretVersion, useUpdateSecret } from '@omnara/react'
 import { ApiError, type Secret } from '@omnara/sdk'
 import { type SyntheticEvent, useState } from 'react'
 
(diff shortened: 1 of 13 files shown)


frontend/apps/web/src/components/org/EditSecretDialog.tsx
```diff
@@ -26,50 +30,134 @@ export function EditSecretDialog({
   orgId: string
   secret: Secret
 }) {
-  const mutation = useUpdateSecret(orgId)
+  const updateSecret = useUpdateSecret(orgId)
+  const createVersion = useCreateSecretVersion(orgId, secret.id)
+  const [savedName, setSavedName] = useState(secret.name)
   const [name, setName] = useState(secret.name)
```
cmsparks: having both the  const [name, setName] and [savedName, setSavedName] feels like a code smell, we can probably consolidate this state, right? (might be easier to do a tanstack form or useReducer instead if we're adding more complex state?)

cmsparks (commented):
lgtm

frontend/apps/web/src/components/org/EditSecretDialog.tsx
asherdale: good call, used `useReducer`

cmsparks (approved):

PR #506, opened 2026-09-10
fix: recover queued agent work and tolerate daemon log failures

asherdale:
- Expire commands queued for five minutes through existing maintenance, complete their waiting tool calls, and let agents resume even when machines appear online or are waking.
- Filter overdue commands from daemon offers and reject late acceptance; preserve execution that was already granted.
- Reject reads of terminal processes that never received execution permission, with a terminal error even when the machine is offline.
- Make supervisor stdout, stderr, and service-log destinations independently best effort so returned write errors do not interrupt child output forwarding or healthy destinations.
- Rename maintenance expiry log messages to cover both queue timeouts and unreachable-machine recovery.

### DIFF
--- a/cmd/maintenance/main.go
+++ b/cmd/maintenance/main.go
@@ -375,12 +375,12 @@ func runCoreMaintenanceTick(
 	if expireDaemonRuntimesErr == nil {
 		expiredDaemonRuntimes = len(records)
 	}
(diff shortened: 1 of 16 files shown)


cmsparks (approved):
lgtm

PR #513, opened 2026-09-11
feat: Implement CIMD oauth for MCP

cmsparks:
Currently, we just support hardcoded API tokens. We can support the CIMD OAuth flow. Adapted the device flow approval for oauth approval for MCP logins.

### DIFF
--- a/docs/api/authentication.mdx
+++ b/docs/api/authentication.mdx
@@ -113,6 +113,39 @@ Keep polling at the returned interval. After `slow_down`, add five seconds to th
 
 A denied or expired flow returns `400` with `"access_denied"` or `"expired_token"`. The PAT does not expire automatically; it appears in the approving user's token list, where it can be revoked, and is used as a bearer token for subsequent API calls.
 
+## Connecting MCP clients
+
+The Omnara API is also exposed as an MCP server at `https://app.omnara.com/api/mcp`. MCP clients that support the [MCP authorization spec](https://modelcontextprotocol.io/specification/draft/basic/authorization) discover the OAuth endpoints themselves and sign you in through the browser, so no token needs to be pasted into the client.
+
+Add the server to your client's `mcp.json`:
+
+```json
+{
+  "mcpServers": {
+    "omnara": {
+      "type": "http",
+      "url": "https://app.omnara.com/api/mcp"
+    }
+  }
+}
+```
+
+On first use the client opens the dashboard's **Authorize access to Omnara** page. Approve it and the client stores its own short-lived tokens, refreshing them in the background.
+
+Clients without OAuth support can pass a personal access token or org API key instead:
+
+```json
+{
+  "mcpServers": {
+    "omnara": {
+      "type": "http",
+      "url": "https://app.omnara.com/api/mcp",
+      "headers": { "Authorization": "Bearer omnara_pat_v1_…" }
+    }
+  }
+}
+```
+
 ## Managing tokens
 
 The **API Tokens** page — or [List personal access tokens](/api-reference/endpoints/users-and-api-keys/list-the-authenticated-users-personal-access-tokens) — shows your tokens with creation and last-used times, metadata only: the secret is never shown again. `last_used_at` tells you which tokens are actually alive; prune the rest, and [revoke](/api-reference/endpoints/users-and-api-keys/revoke-a-personal-access-token) immediately if a token leaks — revocation is permanent.
--- a/internal/httpapi/auth/oauth_authorize.go
+++ b/internal/httpapi/auth/oauth_authorize.go
@@ -1,0 +1,509 @@
+package auth
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/url"
+	"slices"
+	"strings"
+	"unicode/utf8"
+
+	"github.com/omnara-ai/omnara/internal/httpapi/apierror"
+	"github.com/omnara-ai/omnara/internal/httpapi/apimcp"
+	"github.com/omnara-ai/omnara/internal/httpapi/openapi"
+	"github.com/omnara-ai/omnara/internal/resourcename"
+	"github.com/omnara-ai/omnara/internal/storage"
+	"github.com/omnara-ai/omnara/internal/storage/identitystore"
+	"github.com/omnara-ai/omnara/internal/storage/storeerr"
+)
+
+const (
+	OAuthAuthorizePagePath        = "/oauth/authorize"
+	OAuthAuthorizePendingPath     = "/api/auth/oauth/authorize/pending"
+	OAuthAuthorizeApprovePath     = "/api/auth/oauth/authorize/approve"
+	OAuthAuthorizeDenyPath        = "/api/auth/oauth/authorize/deny"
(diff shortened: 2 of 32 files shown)


internal/httpapi/auth/oauth_authorize.go
```diff
@@ -0,0 +1,496 @@
...
+	return ip != nil && ip.IsLoopback()
+}
+
+func (h *Handler) fetchClientMetadata(ctx context.Context, clientID string) (oauthClientMetadata, error) {
+	request, err := http.NewRequestWithContext(ctx, http.MethodGet, clientID, nil)
+	if err != nil {
+		return oauthClientMetadata{}, err
+	}
+	request.Header.Set("Accept", "application/json")
+	response, err := h.clientMetadataHTTPClient.Do(request)
```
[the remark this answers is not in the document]
cmsparks: Inherent in CIMD

internal/httpapi/auth/oauth_authorize.go
```diff
@@ -0,0 +1,507 @@
...
+	if err := validateClientIDURL(clientID); err != nil {
+		writeOAuthError(w, http.StatusBadRequest, "invalid_client", err.Error())
+		return
+	}
+	if form.Get("code") == "" || form.Get("redirect_uri") == "" || form.Get("code_verifier") == "" {
+		writeOAuthError(w, http.StatusBadRequest, "invalid_request", "code, redirect_uri, and code_verifier are required")
+		return
+	}
+	if !h.requireOAuthRateLimits(
+		w, r, "oauth_code_exchange", clientID, TokenConsumeLimit, TokenConsumeLimit, TokenClientLimit, authShortWindow, false,
```
asherdale: this budget is shared by everyone using the same public `client_id`, so ten invalid requests can block other users' valid exchanges/refreshes. should we scope the small budget to the grant while keeping the separate IP limit?

looks like both exchange and refresh are affected

internal/storage/identitystore/password_auth_store.go
```diff
@@ -630,6 +648,12 @@ func (s *Store) RevokeUserAuthTokensTx(ctx context.Context, tx pgx.Tx, userID ID
 	); err != nil {
 		return fmt.Errorf("revoke compromised personal access tokens: %w", err)
 	}
+	if err := qtx.RevokeOAuthAccessTokensForUser(
```
asherdale: should we invalidate outstanding authorization codes here and synchronize approval/exchange with revocation? both existing codes and in-flight approvals can restore access after revoke-all

internal/storage/identitystore/oauth_tokens_store.go
```diff
@@ -0,0 +1,278 @@
...
+		}
+		return OAuthTokenSetRecord{}, storeerr.ErrUnauthorized
+	}
+	if err != nil {
+		return OAuthTokenSetRecord{}, fmt.Errorf("rotate oauth access token: %w", err)
+	}
+	if input.Resource != "" && rotated.Resource != input.Resource {
+		return OAuthTokenSetRecord{}, storeerr.ErrUnauthorized
+	}
+	if err := lockActiveUser(ctx, qtx, rotated.UserID); err != nil {
```
asherdale: can we match revoke-all's lock order here? refresh locks the token before the user, which can deadlock with revocation

internal/storage/queries/identity.sql
```diff
@@ -1327,3 +1327,120 @@ SELECT 'project'::text AS scope, roles.role
...
+  transaction_timestamp() + (sqlc.arg(access_ttl_seconds)::bigint * interval '1 second'),
+  transaction_timestamp() + (sqlc.arg(refresh_ttl_seconds)::bigint * interval '1 second')
+)
+RETURNING id;
+
+-- name: RotateOAuthAccessToken :one
+UPDATE oauth_access_tokens
+SET token_hash = sqlc.arg(token_hash),
+    refresh_token_hash = sqlc.arg(refresh_token_hash),
+    previous_refresh_token_hash = oauth_access_tokens.refresh_token_hash,
```
asherdale: should replaying an older refresh token revoke the grant too? currently, rotating twice makes the original token unrecognizable, so its later reuse won't invalidate the active credentials

internal/storage/identitystore/oauth_tokens_store.go
```diff
@@ -0,0 +1,278 @@
...
+		return OAuthTokenSetRecord{}, storeerr.ErrUnauthorized
+	}
+	if err := lockActiveUser(ctx, qtx, code.UserID); err != nil {
+		return OAuthTokenSetRecord{}, err
+	}
+	tokens, err := newOAuthTokenPair()
+	if err != nil {
+		return OAuthTokenSetRecord{}, err
+	}
+	if _, err := qtx.CreateOAuthAccessToken(ctx, dbsqlc.CreateOAuthAccessTokenParams{
```
asherdale: would it make sense to expose these grants with an individual revoke action, so users can disconnect one MCP client without having to revoke everything?

docs/api/authentication.mdx
```diff
@@ -113,6 +113,39 @@ Keep polling at the returned interval. After `slow_down`, add five seconds to th
 
 A denied or expired flow returns `400` with `"access_denied"` or `"expired_token"`. The PAT does not expire automatically; it appears in the approving user's token list, where it can be revoked, and is used as a bearer token for subsequent API calls.
 
+## Connecting MCP clients
+
+The Omnara API is also exposed as an MCP server at `https://app.omnara.com/api/mcp`. MCP clients that support the [MCP authorization spec](https://modelcontextprotocol.io/specification/draft/basic/authorization) discover the OAuth endpoints themselves and sign you in through the browser, so no token needs to be pasted into the client.
```
asherdale: can the MCP server URL here point to `api.omnara.com` to match the other API endpoints in the docs? also in the two examples below

internal/httpapi/auth/oauth_authorize.go
asherdale: this budget is shared by everyone using the same public `client_id`, so ten invalid requests can block other users' valid exchanges/refreshes. should we scope the small budget to the grant while keeping the separate IP limit?

looks like both exchange and refresh are affected

internal/storage/queries/identity.sql
asherdale: should we still detect reuse after two rotations? an attacker with a stolen refresh token can rotate twice, after which the legitimate client's next refresh fails without revoking the attacker's access

internal/storage/identitystore/password_auth_store.go
asherdale, line 651: should we invalidate outstanding authorization codes here and synchronize approval/exchange with revocation? both existing codes and in-flight approvals can restore access after revoke-all

asherdale (approved):
LGTM

internal/httpapi/auth/oauth_authorize.go
cmsparks: Ah yeah good point, fixing this

internal/storage/identitystore/oauth_tokens_store.go
cmsparks: Yeah, i'll match this by adding some consistent lock helpers

PR #515, opened 2026-09-11
fix: retry browser-session authentication after safe connection failures

asherdale:
- Recover browser-session authentication from stale pooled connections with up to three total attempts and 25/50 ms backoff, retrying only when pgx confirms replay is safe.
- Stop retries when the request ends, classify canceled socket operations and retry waits as client cancellations, and avoid writing a 503 response for those cancellations.
- Preserve genuine connection-acquisition failures and keep partial writes ineligible for retry.

### DIFF
--- a/internal/storage/identitystore/browser_sessions_store.go
+++ b/internal/storage/identitystore/browser_sessions_store.go
@@ -78,7 +78,20 @@ func (s *Store) AuthenticateBrowserSession(
 		TouchIntervalSeconds: int64(browserSessionTouchInterval / time.Second),
 	}
 	row, err := s.q.AuthenticateBrowserSession(ctx, params)
-	if err != nil && ctx.Err() == nil && pgconn.SafeToRetry(err) {
+	for _, delay := range [...]time.Duration{25 * time.Millisecond, 50 * time.Millisecond} {
+		if err == nil || ctx.Err() != nil || !pgconn.SafeToRetry(err) {
+			break
+		}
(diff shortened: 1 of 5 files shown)


internal/storage/identitystore/browser_sessions_store.go
```diff
@@ -70,18 +72,25 @@ func (s *Store) AuthenticateBrowserSession(
...
-	)
+	params := dbsqlc.AuthenticateBrowserSessionParams{
+		TokenHash:            HashBearerToken(token),
+		IdleTimeoutSeconds:   int64(browserSessionIdleDuration / time.Second),
+		TouchIntervalSeconds: int64(browserSessionTouchInterval / time.Second),
+	}
+	row, err := s.q.AuthenticateBrowserSession(ctx, params)
+	if err != nil && ctx.Err() == nil && pgconn.SafeToRetry(err) {
+		row, err = s.q.AuthenticateBrowserSession(ctx, params)
+	}
```
cmsparks, line 96: should we do a retry loop here with backoff instead of just two fetches one after the other?

cmsparks (approved):
lgtm, one comment. if there's not a reason to do the loop, we don't have to, but my intuition is that we should for ephemeral failures like this

internal/storage/identitystore/browser_sessions_store.go
asherdale: good call, added a bounded loop

PR #523, opened 2026-09-14
refactor: normalize agent tools from config sources

asherdale:
- Add missing machine and skill tools to saved YAML/JSON during compilation, preserving explicit disables and permission overrides; keep Slack's contextual tool behavior unchanged.
- Share default-tool rules between compilation and a source-only tool-resolution endpoint used by the builder.
- Show source-added tools under Tools → Other tools with descriptions, permission controls, and explicit disabling; remove associated entries when their last source is removed in the builder.
- Preselect a project-granted cluster-managed pool for new agents only, without falling back to tenant-managed pools.
- Remove obsolete machine-tool warnings and keep shared or merged YAML in the YAML editor to preserve its source safely.


ksarangmath (commented):
~~The new tools look good to me, only thing is I think we can probably do a hard cutover to the new tools and migrate old agents to it. Is there a reason to keep both right now?~~

this was for wrong PR^

frontend/apps/web/src/components/agents/AgentConfigToolPreview.test.tsx
```diff
@@ -0,0 +1,808 @@
...
+      const request = schemas.zResolveAgentConfigToolsRequest.parse(body)
+      const source = z
+        .object({
+          tools: z.record(z.string(), z.object({ enabled: z.boolean().optional() })),
+          mcp: z.record(z.string(), z.unknown()),
+        })
+        .parse(JSON.parse(request.source))
+      const tools = new Map(Object.entries(source.tools))
+      const names: string[] = []
+      if (request.source.includes('machine_pool_name'))
```
ksarangmath: This mock recreates the backend’s machine, skill, subagent, and retrieval defaulting rules. That gives us another implementation to maintain without testing the real one. Could the frontend tests concentrate on how the editor displays and edits the returned preview?

migrations/000038_explicit_default_tools_test.go
```diff
@@ -0,0 +1,263 @@
...
+			require.NoError(t, err)
+			require.True(t, changed)
+			require.Equal(t, before.source, after.source)
+			require.Equal(t, before.sourceHash, after.sourceHash)
+			require.Equal(t, before.id, after.id)
+			require.Equal(t, current.CanonicalJSON, after.compiledDefinition)
+			require.Equal(t, after.compiledDefinition, after.definition)
+			require.Equal(t, current.Hash, after.effectiveDefinitionHash)
+			require.Equal(t, hashBytes([]byte(after.source)), after.sourceHash)
+			recompiled, err := agentconfig.Compile(source.format, []byte(after.source), opts)
```
ksarangmath: There’s some test cleanup available after the source-preservation change: recompiling identical source, checking comments after asserting exact source equality, and repeating subagent model validation already covered elsewhere. The frontend permission tests also overlap and check incidental details like the absence of “Reset.” These could be consolidated around the behavior we need to preserve.

internal/agentconfig/subagent_source.go
```diff
@@ -19,6 +19,14 @@ func SubagentSourceFrom(base AgentConfigSource, subagent SubagentCompiled, depth
 		child.Instruction = strings.TrimSpace(base.Instruction) + "\n\n" + subagent.InstructionAppend
 	}
 	if !depth.CanSpawn() {
+		if len(base.Subagents) > 0 {
+			if child.Tools == nil {
+				child.Tools = make(map[string]AgentConfigToolSource)
+			}
+			for _, name := range missingDefaultToolNames(base) {
```
ksarangmath: Could we simplify subagent configs by keeping the existing compiled-config derivation and removing the separate generated-source path?

Currently, spawning a child does two separate transformations:

Take the parent/profile’s saved compiled config, apply the child’s overrides and depth restrictions, and save the compiled JSON the child actually runs.
Take the parent/profile’s original YAML/JSON, separately apply the child’s changes, and save generated YAML for the editor.
The generated YAML isn’t used to produce the child’s compiled config. These are independently constructed representations of the same child, so they can disagree. This new default-expansion logic adds more behavior to that second path.

The flow I’d prefer is: saved compiled config → apply subagent rules once → validate → save child compiled JSON. The first path already does the core derivation, so that logic and the existing machine-sharing behavior can stay. We’d remove the parallel source generation and allow derived configs to exist without authored source. Parent/profile configs would still retain their original YAML/JSON.

The API/UI support could be a follow-up: return the child’s stored compiled JSON and display that actual configuration. That view doesn’t exist today—the editor depends on source—so we’d need to coordinate removing generated source with that change and decide how editing derived configs should work.

This can be a nonblocking follow-up, but it would eliminate the need to keep generated child source synchronized with what the child actually runs.

ksarangmath: 1. Start from the profile’s saved compiled config. Apply the child’s instruction append and model overrides. A self subagent starts from the parent’s compiled config instead.

2. Use the parent’s currently attached machines. The child shares those existing machines; it doesn’t provision another set from either profile’s pool configuration. Sharing machines also shouldn’t automatically grant additional tool permissions.

3. Depth restricts spawning; it doesn’t grant it.

- Below the depth limit: preserve spawn_agent if the base config enables it.

- At the depth limit: remove spawn_agent and the child’s subagent definitions.

- Keep the other management tools, such as read_agent and send_agent_message, according to their saved permissions.

frontend/apps/web/src/components/agents/AgentConfigToolPreview.test.tsx
asherdale: done, replaced the duplicated backend defaulting logic with fixed preview responses. the tests now focus on how the builder displays and edits those responses

migrations/000038_explicit_default_tools_test.go
asherdale: consolidated the overlapping permission tests and removed redundant source-preservation checks, duplicate model-validation coverage, and incidental UI assertions

internal/agentconfig/subagent_source.go
asherdale: removed the parallel generated-source path. children now derive directly from the saved compiled config, apply instruction/model overrides and depth restrictions, and persist without authored YAML/JSON.

profile children start from the profile’s saved compiled config; self children start from the parent’s. existing machine sharing stays unchanged: children share the parent’s attached machines without provisioning another set or gaining tool permissions.

below the depth limit, saved tool settings are preserved, including absent or explicitly disabled `spawn_agent`. at the limit, `spawn_agent` and subagent definitions are removed while other management tools retain their saved permissions.

migration 39 allows source metadata to be NULL together while preserving config deduplication. pool lookup also supports source-less configs without changing authored configs’ captured-name behavior.

subagent config edits are rejected by the backend, and the frontend shows a read-only notice for both existing and new children. the compiled-config API/viewer is deferred for a follow-up PR as suggested.

internal/storage/executionstore/pool_machines_store.go
```diff
@@ -490,9 +490,13 @@ func listMachinePoolSources(
 		if err != nil {
 			return nil, err
 		}
+		name := grant.PoolName
+		if config.Source != "" {
+			name = configSource.MachineSources[source.Index].MachinePoolName
```
ksarangmath: Could we remove any runtime logic that relies on the yaml/json source? ideally we only rely on the compiled definition for runtime logic

internal/agentconfigcompile/compile.go
```diff
@@ -273,40 +273,15 @@ func DeriveSubagentConfig(
 	if err != nil || configuredModelID == uuid.Nil {
 		return Body{}, fmt.Errorf("subagent model must resolve to a configured project-granted model")
 	}
-	source, sourceFormat, err := deriveSubagentSource(base, subagent, depth)
-	if err != nil {
-		return Body{}, err
-	}
 	return Body{
 		Definition:         json.RawMessage(encoded.CanonicalJSON),
```
ksarangmath: not related to this PR but just noticed while reviewing, I think Definition might just be a duplicate of CompiledDefinition, we could remove Definition completely

ksarangmath (commented):
LGTM, just double check that we don't use the source yaml/json for runtime capability anywhere else in the code, I think the place I commented is the only place

ksarangmath (approved):

internal/agentconfigcompile/compile.go
asherdale: removed `Definition` from application models and reads, `CompiledDefinition` is now the sole application representation. keeping the DB column populated for compatibility this release. next release, i plan to remove the compatibility write and drop the column, sequencing the rollout so no running binaries still reference it

internal/storage/executionstore/pool_machines_store.go
asherdale: makes sense, removed runtime source parsing

asherdale: and yes - i double checked that there are no other cases of using the source yaml/json for runtime capability in the repo

PR #525, opened 2026-09-15
feat: show model api usage

cmsparks:
* Add usage aggregate endpoints for org, project, profile, and agent
* Add frontend UI which displays the usage information
* Add pricing to model catalog
* Show model catalog pricing (i.e. $ / M in + out)


cmsparks:
- [x] Add cost per agent in the table

internal/storage/queries/model_usage.sql
```diff
@@ -0,0 +1,41 @@
...
+  AND revision.id = context.configured_model_revision_id
+JOIN configured_models configured_model ON configured_model.org_id = revision.org_id
+  AND configured_model.id = revision.configured_model_id
+JOIN model_provider_configs provider_config ON provider_config.org_id = revision.org_id
+  AND provider_config.id = revision.model_provider_config_id
+WHERE context.org_id = sqlc.arg(org_id)
+  AND (sqlc.narg(project_id)::uuid IS NULL OR context.project_id = sqlc.narg(project_id)::uuid)
+  AND (
+    sqlc.narg(agent_profile_id)::uuid IS NULL
+    OR (agent.agent_profile_id = sqlc.narg(agent_profile_id)::uuid AND agent.parent_agent_id IS NULL)
```
asherdale: can we add an `include_subagents` option to profile usage like agent usage has? that would let users see the total usage and reported cost of agents launched from the profile, including all their descendants

it seems worthwhile to have if we're already offering this option for a single agent, especially for cases where an agent_profile ends up delegating most of its work to its subagents

frontend/packages/react/src/domains/model-providers.ts
```diff
@@ -70,6 +73,52 @@ export function useModelCatalog(
...
+  const client = useOmnaraClient()
+  const providersQuery = useModelProviders(orgID, { pageSize: 100 })
+  const { fetchNextPage, hasNextPage, isError, isFetching } = providersQuery
+  useEffect(() => {
+    if (!hasNextPage || isFetching || isError) return
+    void fetchNextPage()
+  }, [fetchNextPage, hasNextPage, isError, isFetching])
+  const clusterProviders = (providersQuery.data?.pages ?? [])
+    .flatMap((page) => page.data)
+    .filter((provider) => provider.management_kind === 'cluster')
```
asherdale, line 96: should we show pricing for non-cluster OpenRouter providers too? if not, should we gate the pricing column on the OMNARA_BILLING_URL env var (if we do, we should potentially gate the Usage views on that too)?

frontend/apps/web/src/components/usage/UsageReport.tsx
```diff
@@ -0,0 +1,150 @@
...
+              <TableCell className="text-right tabular-nums">
+                {formatCount(row.tokens.cache_write_input_tokens)}
+              </TableCell>
+              <TableCell className="text-right tabular-nums">
+                {formatCount(row.tokens.output_tokens_total)}
+              </TableCell>
+              <TableCell className="text-right tabular-nums">
+                {formatCount(row.tokens.reasoning_output_tokens)}
+              </TableCell>
+              <TableCell className="text-right tabular-nums">
```
asherdale: can we give the cost header and cells a bit more right padding? the last column currently feels cramped against the table border

Image

frontend/apps/web/src/components/usage/UsageReport.tsx
```diff
@@ -0,0 +1,150 @@
...
+  )
+}
+
+function UsageCostCell({ totals }: { totals: ModelUsageTotals }) {
+  const missing = totals.model_calls - totals.cost.model_calls_with_reported_cost
+  return (
+    <span
+      title={
+        missing > 0
+          ? `${formatCount(missing)} of ${formatCount(totals.model_calls)} calls reported no cost`
```
asherdale: i can't get this explanation to appear when hovering, so the asterisk looks unexplained. can we use our tooltip component instead of the native `title` attribute?

frontend/apps/web/src/components/usage/UsageReport.tsx
```diff
@@ -0,0 +1,150 @@
...
+  TableBody,
+  TableCell,
+  TableHead,
+  TableHeader,
+  TableRow,
+} from '@/components/ui/table'
+import { formatCount, formatUsd } from '@/lib/format'
+import { errorMessage } from '@/lib/submit-status'
+
+export function UsageReportView({
```
asherdale, line 19: right now we only show the all-time total, but users might want to see more information in one place (rather than needing to go to each projects' usage page individually to see the total for each one). could it be worth it to provide a breakdown of the usage based on the parent resource? for example - for org, we could show usage per project. for projects, we could show usage per profile. for profile, we could show usage per agent. this could just be an extra tab/option on each usage view, such as having 'Total' and 'Per Project' for orgs, etc.

in the same vein, would it be helpful to have a date-range picker for this page? that way users could see usage totals for the past week or the past month, instead of being restricted to only all-time stats

feel free to tackle these in a follow-up PR if this is a bigger lift ofc

frontend/packages/react/src/domains/model-providers.ts
```diff
@@ -70,6 +73,52 @@ export function useModelCatalog(
...
+  pricingFor: (
+    modelProviderConfigID: string,
+    providerModelSlug: string,
+  ) => DiscoveredModelPricing | undefined
+  isPending: boolean
+}
+
+export function useClusterModelPricing(orgID: string): ModelPricingLookup {
+  const client = useOmnaraClient()
+  const providersQuery = useModelProviders(orgID, { pageSize: 100 })
```
asherdale, line 88: this pricing lookup requires org-admin access, but the model picker and project models page are available to non-admin project members. those users get a 403 and silently see no prices. project usage already exposes reported costs to project readers, so making granted-model pricing available to them seems consistent. can we either support that access or, if pricing is intentionally admin-only, hide the pricing UI for those users?

frontend/apps/web/src/components/agents/AgentSidebar.tsx
```diff
@@ -220,6 +223,115 @@ function AgentSubagentsGroup({
...
+        ) : (
+          <Collapsible open={expanded} onOpenChange={setExpanded}>
+            <CollapsibleTrigger asChild>
+              <button
+                type="button"
+                className="hover:text-foreground flex w-full items-center justify-between gap-2 py-1.5 text-left text-sm"
+                aria-label={expanded ? 'Hide usage details' : 'Show usage details'}
+              >
+                <span className="truncate tabular-nums">
+                  {formatUsd(query.data.totals.cost.provider_reported_usd)}
```
asherdale: can we distinguish unknown costs here? if no calls report a cost (like for BYOK), this shows `$0.00`, which looks like the usage was free. we could either hide the cost if its 0, or show something like unknown/unavailable?

Image

asherdale (approved):
lgtm 🙌

frontend/packages/react/src/domains/model-providers.ts
cmsparks: I'm going to leave it out, just because we don't know if the user is using BYOK or something weird with openrouter

PR #526, opened 2026-09-15
refactor: use public machine IDs and remove UUID aliases

ksarangmath:
## Summary

Replace agent-specific `machine_ref` selectors (`mchr-...`) with existing public machine IDs (`mch_...`) across machine, command, and artifact tools, approval displays, skill results, and API responses. Parent agents and subagents sharing a machine use the same public ID while retaining separate bindings and permissions. Subagent addressing already uses public `agt_...` IDs.

Remove ref generation and application usage, make the legacy column nullable in migration 000036, and regenerate SQL, OpenAPI, SDK types, and documentation. The removed API response field is recorded as an intentional compatibility exception.

Require canonical lowercase public IDs in the shared decoder. Decode machine selectors at the tool boundary and carry UUIDs through execution; reject noncanonical, padded, and explicitly blank selectors before approval.

Use `uuid.UUID` and `uuid.Nil` directly throughout production code and tests. Remove redundant ID aliases, parsing wrappers, nil-check helpers, and re-exports. Consolidate duplicate nullable conversions on the existing `storeutil` implementations while preserving database and serialization behavior.


migrations/000036_remove_machine_ref.sql
```diff
@@ -0,0 +1,3 @@
+-- +goose Up
+
+ALTER TABLE agent_machine_bindings DROP COLUMN machine_ref;
```
cmsparks: I think this needs to be in a separate PR after we release all the other API / code changes to ensure old pods aren't using the machine_ref column

cmsparks (approved):

migrations/000036_remove_machine_ref.sql
ksarangmath: I'll change this to make it nullable, and then after release i'll merge in a drop

PR #529, opened 2026-09-15
feat: Add docs links and add event log view to chat

cmsparks:
Some mostly minor frontend impovements


frontend/apps/web/src/routes/AgentView.tsx
```diff
@@ -226,7 +265,7 @@ function AgentDock({
           canOperate={canOperate}
         />
       )}
-      {archived ? (
+      {!composer ? undefined : archived ? (
```
asherdale: switching chat → event log → chat drops the draft and attachments because this unmounts `AgentComposer`. can we preserve that state across tabs?

frontend/apps/web/src/components/agents/AgentEventLog.tsx
```diff
@@ -0,0 +1,185 @@
...
+        {showSendInputGuide ? (
+          <SendInputGuide orgId={orgId} projectId={projectId} agentId={agentId} />
+        ) : (
+          <DataTable
+            columns={columns}
+            data={events}
+            getRowId={(event) => event.id}
+            rowExpanded={(event) => <EventJson event={event} />}
+            isPending={chat.historyStatus === 'pending'}
+            isError={chat.historyStatus === 'error'}
```
asherdale: an SSE error stops the stream but leaves `historyStatus` successful, so this view silently shows stale events. should we surface `chat.error` and offer a reconnect/reload action?

frontend/apps/web/src/components/agents/AgentEventLog.tsx
```diff
@@ -0,0 +1,185 @@
...
+        className="text-muted-foreground hover:text-foreground inline-flex items-center gap-1 self-end text-sm transition-colors"
+      >
+        Documentation
+        <ArrowUpRight className="size-3.5" aria-hidden="true" />
+      </a>
+    </div>
+  )
+}
+
+export function AgentEventLog({
```
asherdale, line 148: two suggestions that we already discussed:
- have similar ordering and auto-scroll behavior as the chat tab
- in the empty state for the event log view, have a button to the chat tab that is more visible (something like 'or send a message via the chat' right beneath the code snippets)

frontend/packages/react/src/domains/agent-chat.ts
```diff
@@ -478,12 +480,14 @@ export function useAgentChat(scope: AgentChatScope, options: AgentChatOptions):
...
 
   return {
+    events: data.events,
     messages: projected.messages,
     status: projected.status,
     isWorking: projected.isWorking,
     error: data.error,
     historyStatus: history.status,
     historyError: history.error,
+    retryHistory: () => void history.refetch(),
```
asherdale, line 505: retrying a failed older-page load refetches history that now includes events already in `sessionData.events`. concatenating them duplicates rows and assistant text. should we dedupe by event id and sort by sequence before projecting?

asherdale (approved):
LGTM

PR #531, opened 2026-09-15
feat: offload oversized tool results to artifacts

asherdale:
- Save tool results over 50 KiB from MCP, custom, web, and skill tools as artifacts, returning a bounded preview and file path.
- Add `read_file` and `search_files` for paged text reads and regex searches without an attached machine.
- Make retrieval tools implicitly available while respecting explicit tool configuration and permissions.
- Bound `ask_question` responses and paginate `list_machines`; cap machine descriptions and working directories.


internal/harness/tools/file_read_tools.go
```diff
@@ -0,0 +1,425 @@
...
+type searchFilesRequest struct {
+	Path         string `json:"path"`
+	Pattern      string `json:"pattern"`
+	OffsetLine   *int   `json:"offset_line,omitempty"`
+	MaxMatches   *int   `json:"max_matches,omitempty"`
+	ContextLines int    `json:"context_lines,omitempty"`
+}
+
+func validateReadFileInput(raw json.RawMessage) error {
+	_, _, err := resolveReadFileRequest(raw)
```
cmsparks: This feels like the incorrect way for us to validate, it looks like it's just executing the tool to validate

internal/harness/tools/file_read_tools.go
```diff
@@ -0,0 +1,425 @@
...
+		if *input.OffsetLine < 1 || *input.LimitLines < 1 ||
+			*input.LimitLines > toolcatalog.ReadFileMaxLines {
+			return readFileRequest{}, uuid.Nil, errors.New("invalid artifact line range")
+		}
+	}
+	return input, id, nil
+}
+
+func validateSearchFilesInput(raw json.RawMessage) error {
+	_, _, _, err := resolveSearchFilesRequest(raw)
```
cmsparks: same with above

internal/harness/tools/file_read_tools.go
cmsparks: Generally, the validation should just be schema validation IMO, not really any server side validation. if the tool call fails, it will eventually be caught by the request itself

docs/tools/built-in.mdx
```diff
@@ -180,6 +184,48 @@ Set `path` to `/artifacts/<artifact_id>` and `destination` to the exact machine
...
+
+Reads a UTF-8 text file at `/artifacts/<artifact_id>`. The artifact must contain no NUL bytes and be at most 48 MiB. Readability is determined from the file's bytes, regardless of its declared content type.
+
+| Input | Type | Notes |
+| --- | --- | --- |
+| `path`<span style={{ color: "#c97a7a", fontWeight: 700 }}>*</span> | string | Exact artifact path |
+| `offset_line` | integer | ≥ 1; default 1 |
+| `limit_lines` | integer | 1–200; default 200 |
+| `offset_byte` | integer | ≥ 0; default 0 in byte mode; must be a UTF-8 character boundary |
+| `limit_bytes` | integer | 1–4096; default 4096 in byte mode; use at least 4 to accommodate any character |
```
cmsparks: should we default to the maximum in byte + line mode? I think we should default to a more conservative amount to start

internal/harness/tools/file_read_tools.go
```diff
@@ -0,0 +1,425 @@
...
+func readFileBytes(
+	content []byte,
+	path string,
+	offset, limit int,
+) (map[string]any, error) {
+	if offset > len(content) {
+		offset = len(content)
+	}
+	if offset < len(content) && !utf8.RuneStart(content[offset]) {
+		return nil, errors.New("offset_byte must point to a UTF-8 character boundary")
```
cmsparks: I think the tool call can fail because utf-8 character boundaries are variable width encoding. Consider a string of UTF with byte lengths:

1 1 1 4

Where the agent reads starting at 4 bytes, this tool will likely fail 3+ times in a row

read(4) -> fails
read(5) -> fails
read(6) -> fails
read(7) -> succeeds

The offset + limit should either: snap to the next valid character or be in reference to valid UTF characters to read. I lean towards the latter, which i think is more semantically correct.

An agent will usually start at offset 0, so this wouldn't be an issue, but I think the possibility for the tool call to fail due to rune offsets feels smelly to me.

btw: Claude will complain that utf8.RuneCount is quadratic, but that is a stupid criticism because indexing into a UTF-8 string is almost certainly one extremely well optimized.

cmsparks (approved):
lgtm, just one comment on the tool for read

docs/tools/built-in.mdx
asherdale: ya i like that, lowered the defaults to 100 lines or 512 characters

internal/harness/tools/file_read_tools.go
asherdale: yep you're definitely right here. changed this tool to accept character-based offsets and limits (`offset_char` / `limit_chars`) instead of bytes

PR #532, opened 2026-09-15
fix: issues where MCP returns a 500 instead of a useful error code

cmsparks:


internal/httpapi/mcp_server_tools_routes.go
```diff
@@ -191,13 +191,11 @@ func (s strictOpenAPIServer) mcpServerToolsFailure(
...
 			"auth.secret_id is not available to the project",
 		).WithCause(err)
-	case errors.Is(err, context.Canceled):
-		return nil, apierror.FromCode(openapi.ErrorCodeUpstreamError, "mcp server request was canceled").WithCause(err)
 	case hasStatus && (status == http.StatusUnauthorized || status == http.StatusForbidden):
 		return s.mcpServerAuthRequired(ctx, endpoint, message)
 	default:
 		logpkg.LoggerFromContext(ctx).WarnContext(ctx, "mcp tool discovery failed", "error", err)
-		return nil, apierror.FromCode(openapi.ErrorCodeUpstreamError, message).WithCause(err)
+		return mcpServerUnreachableResponse(message), nil
```
ksarangmath, line 214: Errors reaching this branch can come from credential loading, OAuth refresh, or catalog storage as well as the MCP server. I reproduced successful discovery followed by a failed catalog write returning 422. We should preserve the distinction between configuration problems, temporary upstream failures, and failures inside Omnara

internal/httpapi/mcp_oauth_routes.go
```diff
@@ -277,8 +277,8 @@ func (s *Server) resolveMCPOAuthClientForAPI(
 		)
 		if err != nil {
 			apiErr := apierror.FromCode(
-				openapi.ErrorCodeUpstreamError,
-				"mcp client registration failed: "+err.Error(),
+				openapi.ErrorCodeUnprocessable,
```
ksarangmath: A registration endpoint returning a temporary 503 now becomes a 422 telling the user to supply a client ID. That doesn’t establish that registration was rejected or that their configuration needs changing. Our API docs tell callers to retry 5xx but fix the request for 422. Temporary failures should remain distinguishable here, in OAuth discovery, and in tool discovery.

internal/httpapi/mcp_server_tools_routes.go
```diff
@@ -191,13 +191,11 @@ func (s strictOpenAPIServer) mcpServerToolsFailure(
 			openapi.ErrorCodeNotFound,
 			"auth.secret_id is not available to the project",
 		).WithCause(err)
-	case errors.Is(err, context.Canceled):
```
ksarangmath, line 194: nit: Returning a response with a nil error loses that cause, so cancellations that reach this branch now appear as 422 warnings instead of 499 at info level.

api/openapi/openapi.yaml
```diff
@@ -8447,7 +8436,7 @@ components:
     MCPServerAuthRequiredError:
       type: object
       additionalProperties: false
-      required: [error, code, auth]
+      required: [error, code]
```
ksarangmath: This schema now covers discovery failures where authentication isn’t the issue, but it’s still named MCPServerAuthRequiredError. Could the name reflect that broader meaning?

api/openapi/openapi.yaml
```diff
@@ -2440,13 +2439,11 @@ paths:
         '404':
           $ref: '#/components/responses/NotFound'
         '422':
-          description: The MCP server rejected the configured authentication. `auth` hints which auth type the server expects.
+          description: The MCP server could not be reached, did not speak MCP, or rejected the configured authentication. When the server rejected authentication, `auth` hints which auth type it expects.
```
ksarangmath: nit: The operation description at line 2415 still says a rejected connection produces 422 with an auth hint. The auth probe can now fail and return 422 without that hint. Could the description reflect that distinction too?

frontend/apps/web/src/components/agents/AgentConfigMcpServerTools.tsx
```diff
@@ -325,7 +325,7 @@ function discoveryFailureTitle(
       ? 'The server rejected the selected secret.'
       : 'This server expects a bearer token. Switch authentication to a bearer secret.'
   }
-  if (cause instanceof ApiError && cause.status === 502) {
+  if (cause instanceof ApiError && cause.status === 422) {
```
ksarangmath: A 422 without an auth hint doesn’t necessarily mean we couldn’t connect. We can reach the server and then fail while reading its MCP response or discovering tools. This title seems more specific than the error actually establishes.

tools/ci/openapi-compat/approved-breaking-changes.txt
```diff
@@ -51,3 +51,4 @@ Strict model resource creation: POST /orgs/{orgID}/model-provider-configs remove
 Strict model resource creation: POST /orgs/{orgID}/model-provider-configs/{modelProviderConfigID}/models removed the success response with the status `200`
 Strict model resource creation: POST /orgs/{orgID}/projects/{projectID}/model-grants removed the success response with the status `200`
 Public machine IDs replace binding refs: POST /orgs/{orgID}/projects/{projectID}/agents removed the required property `machine_bindings/items/machine_ref` from the response with the `201` status
+MCP discovery errors are client-facing 422s: POST /orgs/{orgID}/projects/{projectID}/mcp-servers/tools the response property `auth` became optional for the status `422`
```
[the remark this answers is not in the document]
cmsparks: Verified against `origin/main` by running `oasdiff breaking` with the allowlist removed entirely: it reports exactly one breaking change, the `auth` property becoming optional on the tools `422`, which is the line already added. Removing the `502` responses and the `upstream_error` enum value are not flagged under the configured severity levels, so no additional allowlist entries are needed. `make openapi-compat-check COMPAT_BASE_SHA=origin/main` passes.

api/openapi/openapi.yaml
cmsparks: Fixed in dcfefe13f.

internal/httpapi/mcp_server_tools_routes.go
cmsparks: Fixed in dcfefe13f.

internal/httpapi/mcp_oauth_routes.go
cmsparks: Fixed in b2de4c86f. Transient upstream failures (5xx, 429, timeouts, connection errors) from the MCP server or its authorization server now return 424 with code `upstream_unavailable`, documented as retryable. Definitive rejections (registration 400 with an error body, non-MCP responses, missing metadata) keep returning 422. Applied to OAuth discovery, dynamic client registration, and tool discovery including the auth probe.

internal/httpapi/mcp_server_tools_routes.go
cmsparks: Fixed in a67b78a76. The MCP manager now wraps errors at their origin with `ErrInternal` (storage and encoding faults inside Omnara), `ErrCredential` (auth secret cannot be used or refreshed), and `ErrRefreshBusy` (lease contention). Tool discovery maps them: internal → 500 logged at error level, credential → 424 when the token endpoint failed transiently or 422 otherwise (no auth hint, since the auth type is already right), busy → 409. A failed catalog write now surfaces as 500 instead of 422.

cmsparks: Classed internal errors as 500, unreachably/flaky is 424, 422 is failure to process request. attached the auth hint in that case

internal/httpapi/mcp_oauth_routes.go
cmsparks: Does a 424 now, which is distinct from 503 and 422

internal/httpapi/mcp_server_tools_routes.go
cmsparks: fixed

api/openapi/openapi.yaml
cmsparks: fixed, clarified all the outcomes in the docs

internal/httpapi/mcp_oauth_routes.go
```diff
@@ -155,10 +155,11 @@ func (s strictOpenAPIServer) startMCPOAuth(
 	defer cancel()
 	requirement, err := mcp.DetectAuth(outboundCtx, mcpURL, mcp.AuthOptions{HTTPClient: s.server.mcpOAuthHTTPClient})
 	if err != nil {
-		apiErr := apierror.FromCode(
-			openapi.ErrorCodeUpstreamError,
-			"mcp authorization discovery failed: "+err.Error(),
-		).WithCause(err)
+		apiErr := mcpUpstreamFailure(
```
ksarangmath, line 158: Some temporary OAuth metadata failures still become 422 here e.g. the MCP server’s protected-resource metadata returning 503, or the authorization server’s metadata returning 429. Tool discovery also interprets these failures as a bearer-token requirement. Can we preserve the distinction between missing metadata and temporarily failing to fetch it, so an upstream outage doesn’t become a configuration or authentication instruction?

internal/mcp/dcr.go
```diff
@@ -0,0 +1,21 @@
+package mcp
+
+import (
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+var clientRegistrationStatusPattern = regexp.MustCompile(`^registration failed with status (\d{3})[^:]*: (.*)$`)
```
ksarangmath, line 9: Is it possible to use the actual HTTP status instead of parsing from a string here? This is kinda brittle, e.g. a newline could cause a 503 error to become a 422 and tell the user to supply client_id. Maybe we can get the status from an http client we pass into the sdk

internal/mcp/catalog.go
ksarangmath, line 362: nit, predated this PR but saw it while reviewing this pr: Starting a new catalog refresh leaves the previous RefreshError in place, so a concurrent discovery request can hit this branch and receive the old failure while the retry is still running. Recreating it from text also loses its error category, so a stored 503 becomes 422 under the new mapping. Can concurrent callers follow the current refresh state instead of receiving a previous attempt’s untyped error?

internal/mcp/errors.go
```diff
@@ -179,6 +190,11 @@ func IsRetryableConnectionFailure(cause error) bool {
 	if errors.As(cause, &netErr) && netErr.Timeout() {
 		return true
 	}
+	var opErr *net.OpError
+	var dnsErr *net.DNSError
+	if errors.As(cause, &opErr) || errors.As(cause, &dnsErr) || errors.Is(cause, io.ErrUnexpectedEOF) {
```
ksarangmath, line 195: A server closing the connection before sending an HTTP response produces a wrapped io.EOF, which this helper doesn’t recognize, io.ErrUnexpectedEOF is a different error. Tool discovery returns 422 even though the server never responded. Could this remain a retryable upstream-unavailable failure?

internal/mcp/discovery.go
```diff
@@ -79,7 +79,7 @@ func newDiscoveredServer(
 ) (DiscoveredServer, error) {
 	var serverInfo sdkmcp.Implementation
 	if err := json.Unmarshal(serverInfoJSON, &serverInfo); err != nil {
-		return DiscoveredServer{}, fmt.Errorf("decode mcp server info: %w", err)
+		return DiscoveredServer{}, fmt.Errorf("%w: decode mcp server info: %w", ErrInternal, err)
```
ksarangmath, line 82: nit: The upstream MCP response could have caused the error if it doesn't obey MCP protocol correct, e.g. it returns an int for the name key. We could validate the response to see if it's an upstream issue, right now we say it's an internal omnara server error

internal/mcp/catalog.go
```diff
@@ -339,7 +339,8 @@ func (m Manager) refreshCatalogUntil(
 		)
 		if err != nil {
 			return executionstore.MCPServerCatalogRecord{}, fmt.Errorf(
-				"acquire mcp catalog refresh lease for %s: %w",
+				"%w: acquire mcp catalog refresh lease for %s: %w",
+				ErrInternal,
```
ksarangmath, line 343: If this storage call returns context.Canceled because the caller disconnected, we also attach ErrInternal. The request logger then stops recognizing it as cancellation-only and records a 422 warning instead of 499 at info level. Could we keep caller cancellation distinct when tagging storage failures? The same issue applies to wrapStoreErr.

docs/api/overview.mdx
```diff
@@ -84,7 +84,7 @@ Branch on `code` — it's a stable contract, enumerated in the OpenAPI spec, and
 Two rules worth internalizing:
 
 - **Resources you cannot read return `404`, not `403`.** The API doesn't confirm the existence of things outside your access. A `403` always means "the resource is visible to you, but your role doesn't permit this action."
-- **Retry `429` and `5xx` with backoff, reusing the same `Idempotency-Key`** so a retry can never act twice. Other `4xx` errors won't change on retry — fix the request instead.
```
ksarangmath, line 87: very nit: This says other 4xx errors won’t change on retry, but this PR adds a 409 for an MCP catalog or OAuth refresh already in progress, with instructions to retry shortly.

internal/httpapi/mcp_server_tools_routes.go
ksarangmath: nit: An MCP server returning JSON-RPC -32603 (internal error) inside an HTTP 200 response reaches this fallback and becomes 422. That reports a failure inside the upstream server, but we classify it as a request problem.

ksarangmath (commented):
Found a few more cases where error codes might not be indicative of what actually happened

internal/mcp/dcr.go
cmsparks: Yeah we could, that's originally what this code was, but it felt a little unnecessary because the error is from the library, so the error message will never change unless we update the library version.

PR #534, opened 2026-09-16
feat: add custom model providers and link to add models from grant dialog

asherdale:
Closes OMN2-380 and OMN2-381.

- Add a **Custom endpoint** option to the Add model provider dialog with an API format select (OpenAI Responses, OpenAI Chat Completions, Anthropic Messages) and a base URL field. The request sends `api_format` and `base_url` and lets the server default endpoint path and auth from the format. No backend changes.
- Model discovery and the existing failure step already handle non-preset endpoints, so custom providers flow into the same add-models step as presets.
- Add an "Add more models to {provider}" link under the provider picker in the project Grant models dialog, shown only to users who can manage the org since the target page gates the create dialog on that role.
- The link deep-links to `/models?provider=<id>`. The models route gains a typed `provider` search param, and the configured models section opens the create-model dialog with that provider preselected, clearing the param on close.
- `CreateConfiguredModelDialog` accepts an optional `defaultProviderId` to seed the provider field.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


frontend/apps/web/src/components/org/CreateConfiguredModelDialog.tsx
```diff
@@ -40,11 +40,13 @@ export function CreateConfiguredModelDialog({
   onOpenChange,
   orgId,
   providers,
+  defaultProviderId = '',
```
cmsparks: default probably shouldn't be an empty string if its optional?

cmsparks (approved):
lgtm, one typign nit

PR #535, opened 2026-09-16
feat: Tool search and deferred loading

cmsparks:
Implement tool search and deferred loading. The implementation varies by provider:
* Common tool search tool for each provider, which takes a regex query and returns tool definitions
* Anthropic provider maps tool definitions to tool_addition blocks. Tools are marked as deferred when attached
* Openai responses maps tool defintions to tool_search_output blocks. Tools are marked as deferred when attached
* Chat completions maps tool definitions to raw text blocks with schemas output. Chat completions then has an additional special case tool call deferred tool, which takes in an arbitrary json schema in order to preserve cache. this gets mapped transparently to real tool calls later


internal/model/anthropicmessages/request.go
```diff
@@ -233,6 +300,11 @@ func buildMessages(
 					Content:   toolResultContent(result, bundle.ResolvedMedia),
 					IsError:   result.Outcome == executionstore.ToolResultOutcomeFailed,
 				})
+				if search, ok := modelcontext.ToolSearchResultFromToolResult(result); ok {
+					for _, spec := range modelcontext.DiscoveredToolSpecs(bundle.ToolSpecs, search) {
+						pendingToolAdditions = append(pendingToolAdditions, spec.Name)
+					}
```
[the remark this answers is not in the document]
cmsparks: This is false

internal/model/anthropicmessages/request.go
```diff
@@ -136,9 +140,69 @@ type toolDefinition struct {
...
+		names = append(names, definition.Name)
+	}
+	return names
+}
+
+func toolAdditionBlocks(names []string) []any {
+	blocks := make([]any, 0, len(names))
+	for _, name := range names {
+		blocks = append(blocks, toolAdditionBlock{
+			Type: "tool_addition",
```
asherdale: why use the beta `tool_addition` instead of returning `tool_reference` blocks in the search's `tool_result`? standard tool references preserve prompt caching and work on Sonnet 5 and 4.6, which this approach excludes. seems like we could avoid the compatibility restriction and beta header

internal/model/openairesponses/request.go
```diff
@@ -126,26 +127,59 @@ type responsesReasoning struct {
...
+
+func functionToolDefinition(spec modelcontext.ToolSpec) responsesTool {
+	strict := false
+	return responsesTool{
+		Type:         "function",
+		Name:         spec.Name,
+		Description:  spec.Description,
+		Parameters:   toolParameters(spec),
+		Strict:       &strict,
+		DeferLoading: spec.Deferred,
```
asherdale, line 165: can we account for deferred tools in `EstimatePreparedRequest` too? it still counts all the unloaded schemas, so a large catalog can trigger context-budget failures or unnecessary compaction. this affects the Anthropic adapter too

internal/model/openaichatcompletions/request.go
```diff
@@ -344,27 +360,38 @@ func assistantMessagesForEntry(
...
-) []chatMessage {
+	bundle modelcontext.Bundle,
+) ([]chatMessage, error) {
 	var mediaContent []any
+	deferredEnabled := modelcontext.DeferredToolsEnabled(bundle.ToolSpecs)
 	for _, result := range results {
+		content := toolResultOutput(result, bundle.ResolvedMedia)
+		if deferredEnabled {
+			if search, ok := modelcontext.ToolSearchResultFromToolResult(result); ok {
+				output, err := toolSearchOutputContent(search)
```
asherdale: can we use `DeferredToolSearchDefinitions` here too, like the other adapters? Chat Completions currently replays the saved schema, so it’s the only adapter that keeps showing the old arguments after a tool’s schema changes

internal/agentconfig/runtime_contract.go
```diff
@@ -181,12 +182,32 @@ func RuntimeContractFromCompiled(
 			}
 		}
 	}
+	if contract.DefersAnyTool() {
+		contract, err = contract.WithImplicitBuiltInTool(toolcatalog.ToolNameToolSearch)
```
asherdale: now we have a new system for implicit tools, which moves them into compiled config without modifying source YAML/JSON. so after rebasing, here's the easy way to go about updating this:

- move `tool_search` inclusion from `RuntimeContractFromCompiled` into the shared `missingDefaultToolNames` / `compileTools` path, preserving explicit disable/overrides. compilation and the tool-preview endpoint should use the same rule
- mark `tool_search` as `implicit` in the catalog instead of adding a frontend hidden-name exception; the builder uses that flag for “other tools”
- update `useAgentBuilderTools` to forward deferred settings and relevant MCP defaults/per-tool overrides (the current preview payload omits them)
- preserve `enabled: false` when merging the builder extraction changes, and test compiled/preview agreement plus builder round-trips

this is only about where `tool_search` gets added to the config, so no need to change anything with the adapters. happy to take another brief look at this later if you want confirmation that everything looks good, just lmk

internal/toolcatalog/catalog.go
```diff
@@ -79,6 +79,12 @@ const (
 	downloadFileToolDescription = "Copy a file from Omnara's virtual filesystem to an attached machine. " +
 		"Currently supports /artifacts/<artifact_id> as path; provide destination. " +
 		"If the download is still running after the initial wait, use the returned process_id with the process tools."
+	toolSearchToolDescription = "Search the tools that are declared but not loaded into this conversation, " +
+		"and load the matches so they can be called on the next turn. " +
```
asherdale: minor nit: could we say something like "after the search returns" instead of "on the next turn"? it looks like these tools can be called within the same agent turn, so the current wording could imply they need to wait for another user message

asherdale (approved):
lgtm 👌

internal/model/anthropicmessages/request.go
cmsparks: Good point, that's also a bit closer to responses api.

internal/model/openaichatcompletions/deferred_tools.go
```diff
@@ -0,0 +1,95 @@
...
+	return names
+}
+
+func toolSearchOutputContent(specs []modelcontext.ToolSpec, search modelcontext.ToolSearchResult) (string, error) {
+	output := toolSearchOutput{
+		Pattern:            search.Pattern,
+		ToolNames:          search.ToolNames,
+		TotalDeferredTools: search.TotalDeferredTools,
+		Tools:              modelcontext.DeferredToolSearchDefinitions(specs, search),
+	}
```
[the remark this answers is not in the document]
cmsparks: Not relevant, intentional

internal/model/openaichatcompletions/request.go
```diff
@@ -344,27 +360,38 @@ func assistantMessagesForEntry(
...
+			Content:    content,
 		})
-		mediaContent = append(mediaContent, toolResultMediaContent(result, media)...)
+		mediaContent = append(mediaContent, toolResultMediaContent(result, bundle.ResolvedMedia)...)
 	}
 	if len(mediaContent) > 0 {
 		messages = append(messages, chatMessage{Role: chatRoleUser, Content: mediaContent})
 	}
-	return messages
+	return messages, nil
```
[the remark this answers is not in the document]
cmsparks: should be fine, just inconsistency with display

internal/model/anthropicmessages/request.go
asherdale: can we return just the tool references when there are matches? appending them to the existing result text makes Sonnet 5 and 4.6 reject the request with "Tool definitions/code execution functions cannot be mixed with other content". returning only the references works

PR #536, opened 2026-09-16
fix: back off daemon crashes and refresh supervisors on update

asherdale:
- Back off repeated daemon crashes from 3 seconds to a 3-minute maximum, resetting after 5 minutes of stable operation or a manual restart.
- Persist bounded crash output, exit reason, uptime, and peak memory in machine failure reports for later inspection, without requiring writable storage for capture.
- Refresh the foreground supervisor after daemon updates while preserving its PID and runtime lock; retain compatibility with older supervisors and preserve handoff capability during update-failure recovery.


internal/omnarad/supervisor_output.go
```diff
@@ -0,0 +1,78 @@
...
+			}
+			switch string(b.line[:b.lineSize]) {
+			case "panic:", "fatal error:", "runtime: out of memory:",
+				"runtime: goroutine stack exceeds ", "runtime: program exceeds ",
+				"runtime: failed to create new OS thread", "runtime: mmap":
+				b.truncated = b.truncated || b.size+i+1 > b.lineSize
+				b.size = copy(b.data[:], b.line[:b.lineSize])
+				b.fatal = true
+				p = p[i+1:]
+				break scan
```
cmsparks: We should not be using gotos here (but probably resolved by my above comment)

internal/omnarad/supervisor_output.go
```diff
@@ -0,0 +1,78 @@
...
+	line      [len("runtime: failed to create new OS thread")]byte
+	lineSize  int
+}
+
+type supervisorOutputWriter struct {
+	tail   *supervisorOutputTail
+	stderr bool
+}
+
+func (w supervisorOutputWriter) Write(p []byte) (int, error) {
```
cmsparks: This function is kinda hard to read, 3 things:
* could we do less reassignment of temp vars? n := len(p) and b := w.tail obfuscates the meaning and source of these variables
* for scanning, maybe case "panic:", ...: should be some list of constant strings that we loop against? 
* We can probably use some stdlib functions instead of doing a ton of manual string indexing (or at least break things up into smaller helpers). I know we have tests, but it's hard for me to know that its correct and efficient

internal/omnarad/supervisor_output.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/omnara-ai/omnara/internal/machinedaemon"
+)
+
+type supervisorOutputTail struct {
+	mu        sync.Mutex
+	data      [machinedaemon.MaxFailureDetailBytes]byte
+	size      int
+	truncated bool
+	fatal     bool
+	line      [len("runtime: failed to create new OS thread")]byte
```
cmsparks: I don't like this line definiton (especually with the hardcoded len(...)), i think this probably should be a slice, like `[]byte`

internal/omnarad/supervisor.go
```diff
@@ -73,7 +84,11 @@ func runForegroundSupervisor(ctx context.Context, home string, log *slog.Logger)
 	}()
 	childStdout, childStderr := supervisorChildWriters(os.Stdout, os.Stderr, logFile)
 	log = slog.New(logpkg.NewJSONHandler(childStdout, nil))
-	return runSupervisorLoop(ctx, home, daemonRestartDelay, restart, childStdout, childStderr, log)
+	return runSupervisorLoop(ctx, home, supervisorRestartPolicy{
```
cmsparks: i dont think we need this supervisorRestartPolicy type because daemonRestartDelay, daemonRestartMaxDelay and daemonRestartResetAfter are constants (and we only use it here and in tests

cmsparks (commented):
lgtm, mostly nits

cmsparks (approved):

internal/omnarad/supervisor_output.go
asherdale: ya good call - moved capture logic onto the tail, used descriptive names, and pulled markers into one list

internal/omnarad/supervisor_output.go
asherdale: def agree, changed this to []byte

PR #538, opened 2026-09-16
fix: clarify interactive login in setup skill

asherdale:
- Clarify that setup login runs interactively without piping output, using `--no-browser` on headless or remote machines.
- Tell the agent to share the approval link and code, keep login running, and confirm authentication before continuing setup.
- Fixes OMN2-342.


skills/setup-omnara-agent/SKILL.md
```diff
@@ -6,7 +6,12 @@ description: Install and set up Omnara. Use when the user asks to install Omnara
 Read https://docs.omnara.com/llms.txt and the Quickstart guide first. Fetch other
 documentation pages as you need them.
 
-Before doing anything, check if the user is logged in with `npx omnara whoami`. If they are not logged in, run `npx omnara login` and direct the user to the login link.
+Before setting up the agent, check authentication with `npx omnara whoami`. If not
+logged in, run `npx omnara login` in an interactive terminal session without piping
```
ksarangmath, line 10: nit: should we specify to the agent that this will pop up the browser for the user to login interactively, basically it can wait for their user to do it through the browser if their user asks what's goin on or something.

also "immediately share the approval link and code" - does this happen if the agent starts the login interactively?

lastly (if you haven't already) can you test on your mac and a cloud vm to see if both work fine (just copy and paste this skill to the agent)

ksarangmath (approved):
LGTM

skills/setup-omnara-agent/SKILL.md
asherdale: added the browser explanation. yep, interactive login prints the approval URL and code before waiting, so the agent can share both. tested on mac and a fresh cloud sandbox with `--no-browser`; login and the follow-up `whoami` succeeded in both. the cloud agent shared the link/code and kept login running through approval

PR #539, opened 2026-09-17
fix: Implement UI nit fixes

cmsparks:
Small UI nits:
* refresh sub agents list upon subagent tool call
* improve copy and formatting on the subagent config ui
* fix cost ui in config view
* attempt to match prefix for pricing if not valid


frontend/packages/react/src/domains/agent-interactions.ts
```diff
@@ -107,3 +113,34 @@ export function useCancelAgent(orgID: string, projectID: string, agentID: string
...
+}
+
+export function invalidateSubagentList(
+  queryClient: QueryClient,
+  client: OmnaraClient,
+  scope: AgentChatScope,
+  event: AgentEvent,
+  events: AgentEvent[],
+): void {
+  const subagentInput = event.event_kind === 'agent_input' && event.agent_id !== scope.agentID
```
asherdale: subagent replies have the parent's `agent_id`, so this condition never matches real replies. should we use the sender's actor or invalidate on parent inputs instead? the test currently masks this by setting `agent_id` to the child

frontend/apps/web/src/components/agents/AgentConfigModelField.tsx
```diff
@@ -21,18 +22,23 @@ interface ModelChoice extends ConfiguredModelSummary {
...
+  renderItem: (model) => {
+    const row = (
+      <span className="flex min-w-0 flex-1 items-baseline gap-1.5">
+        <span className="truncate">{model.name}</span>
+        <span className="text-muted-foreground truncate text-xs">{model.provider_config}</span>
+      </span>
+    )
+    if (!model.pricing) return row
+    return (
+      <Tooltip>
```
asherdale: it looks like there are two competing tooltips now - when the label is truncated, the existing `OverflowTooltip` opens after 500 ms and replaces this price tooltip

frontend/packages/react/src/domains/agent-interactions.ts
```diff
@@ -107,3 +113,34 @@ export function useCancelAgent(orgID: string, projectID: string, agentID: string
...
+  if (!subagentInput && !spawnsSubagent(event, events)) return
+  void queryClient.invalidateQueries({
+    queryKey: listAgentsQueryKey({
+      path: { orgID: scope.orgID, projectID: scope.projectID },
+      client,
+    }),
+  })
+}
```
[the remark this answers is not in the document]
asherdale: this seems legit - if you open the page while a spawn awaits approval, this lookup misses the call, so the child doesn't appear in the sidebar

asherdale (approved):
LGTM

frontend/packages/react/src/domains/agent-interactions.ts
cmsparks: Yeah, you're right it should be using the actorid

frontend/packages/react/src/domains/model-providers.ts
```diff
@@ -83,6 +84,20 @@ export interface ModelPricingLookup {
...
+
+export function catalogPricing(
+  models: DiscoveredProviderModel[],
+  slug: string,
+): DiscoveredModelPricing | undefined {
+  const exact = models.find((model) => model.slug === slug)
+  if (exact) return exact.pricing
+  const dated = new RegExp(`^${escapeRegExp(slug)}-\\d+$`)
+  return models.find((model) => model.pricing && dated.test(model.slug))?.pricing
+}
```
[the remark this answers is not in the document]
cmsparks: This doesn't affect anything

PR #540, opened 2026-09-17
fix: stop messaging the parent agent when a subagent is archived

cmsparks:


internal/storage/executionstore/subagents_integration_test.go
```diff
@@ -438,7 +438,7 @@ func TestLaunchSubagentRejectsForeignParentAndDepthLimit(t *testing.T) {
 	}
 }
 
-func TestSubagentArchiveNotifiesParent(t *testing.T) {
+func TestSubagentArchiveDoesNotMessageParent(t *testing.T) {
```
ksarangmath, line 441: nit: seems like a negative test that's probably unecessary

internal/storage/executionstore/subagents_store.go
```diff
@@ -30,7 +30,6 @@ const (
 	SubagentMessageKindFailed          = "failed"
 	SubagentMessageKindQuestion        = "question"
 	SubagentMessageKindCanceled        = "canceled"
```
ksarangmath, line 32: Could we also remove any message that's sent during cancellation? ideally the tool call just returns whether it canceled or not, and the agent can use read/list agent tools to see the status, and doesn't need an extra agent input

ksarangmath: also if we remove this, there's a lock that locks the parent because the new input would also be written (I think), should double check whether we need that lock anymore if we remove the new agent input sent to the parent on cancelation

ksarangmath (approved):
LGTM

PR #542, opened 2026-09-17
feat: warn before leaving unsaved agent edits

asherdale:
- warn before closing, reloading, or navigating away from unsaved agent drafts and agent/profile configuration edits
- allow navigation that preserves the editor and avoid duplicate prompts after successful submissions or confirmed launch/delete actions
- suppress warnings during MCP OAuth redirects that preserve the draft, restoring protection when returning via browser history

image


frontend/apps/web/src/hooks/use-unsaved-changes-warning.ts
```diff
@@ -0,0 +1,35 @@
+import { type ShouldBlockFn, useBlocker } from '@tanstack/react-router'
+
+let suppressed = false
```
cmsparks: no module level state, especially when we're using hooks

cmsparks (approved):

PR #544, opened 2026-09-17
fix: stabilize agent builder tool controls

asherdale:
- Keep automatic tool rows visible during preview refreshes and failures, and preserve their order when permissions change.
- Remove regular tool rows immediately when deleted, even if the preview refresh fails.
- Add a compact hover tooltip explaining “Other tools”.
- Hide the built-in Loaded/Deferred dropdown while preserving deferral settings in YAML.


frontend/apps/web/src/components/agents/useAgentBuilderTools.ts
```diff
@@ -45,8 +46,20 @@ export function useAgentBuilderTools(
...
+  if (scopeChanged || (query.data !== undefined && query.data !== lastResult.data)) {
+    setLastResult({ ...scope, data: query.data })
+  }
+  const data = query.data ?? lastResult.data
+  return {
+    data,
+    isPending: query.isPending && data === undefined,
+    isError: query.isError,
+    refetch: query.refetch,
+  }
```
cmsparks: We should be able to fix this in the base useAgentConfigTools hook with initialData or placeholderData function in react query, which handles this type of thing built in. 

In general, we should not be combining setState with react query hooks, that's an antipattern. We should use the higher level features of react query

cmsparks: Here's an example replacement:

```ts
export function useAgentConfigTools(
  orgID: string,
  projectID: string,
  request: ResolveAgentConfigToolsRequest,
) {
  const client = useOmnaraClient()
  const queryClient = useQueryClient()
  const scopeKey = ['agent-config-tools', orgID, projectID] as const
  return useQuery({
    queryKey: [...scopeKey, request],
    queryFn: async ({ signal }) => {
      const { data } = await sdk.resolveAgentConfigTools({
        path: { orgID, projectID },
        body: request,
        client,
        signal,
      })
      return data
    },
    // Fetch the initial placeholder data from cache (effectively merges with previous values), so we don't need to do the imperative setState.
    initialData: () =>
      queryClient
        .getQueryCache()
        .findAll({ queryKey: scopeKey })
        .filter((query) => query.state.data !== undefined)
        .sort((a, b) => b.state.dataUpdatedAt - a.state.dataUpdatedAt)[0]?.state.data,
    initialDataUpdatedAt: 0,
    enabled: orgID !== '' && projectID !== '',
    staleTime: 60 * 1000,
    retry: false,
  })
}


export function useAgentBuilderTools(
  source: BasicConfig,
  scope: { orgId: string; projectId: string },
) {
  return useAgentConfigTools(scope.orgId, scope.projectId, {
    source_format: 'json',
    source: agentBuilderToolsSource(source),
  })
}
```

cmsparks (commented):
have a comment on structuring the react hook

frontend/apps/web/src/components/agents/useAgentBuilderTools.ts
asherdale: went with `initialData` seeding as suggested, with two small additions: the key is scoped per editor with `useId` so a fresh editor isn't seeded from a sibling's stale list, and `queryFn` drops inactive entries that only ever held seeded data, so revisiting a config whose fetch failed shows the last list displayed rather than its old seed

lmk if you'd rather me handle either differently

cmsparks (approved):

PR #545, opened 2026-09-17
fix: discover machine pools through list_machines using stable IDs

ksarangmath:
## Summary

Pool choices are currently injected into each model request, while `create_machine` selects pools by mutable name. This moves discovery into `list_machines` and uses stable pool IDs through creation and approval, so renaming pools cannot redirect a request.

`list_machines` now returns separate `machines` and `machine_pools` collections, including configured pools with no machines. Pool discovery uses the model call's saved compiled config and current project grants. Both collections share bounded pagination, and pool-backed machine results include their pool ID.

`create_machine` accepts `machine_pool_id` instead of `machine_pool_name`; the ID can still be omitted when one pool is available. Approval displays the pool's name and ID but binds only to its identity. Tool enablement remains independent, so users can still disable `list_machines` while enabling creation.

Removes pool-context loading, validation, and injection from the shared context builder and all three provider adapters, along with the obsolete context tests. Updates the tool catalog and documentation.

## Validation

- Affected Go unit suites passed: harness, model providers, model context, tool catalog, execution storage, and config compilation.
- Six focused database integration tests passed using an isolated PostgreSQL instance. Coverage includes source-less config discovery, empty pools, name swaps before approval, renaming after approval, creation, list/inspect pool IDs, revoked grants, and BYO grant handling.
- Pagination tests cover both collections, encoded size boundaries, deleted cursor resources, and continuation past oversized entries.
- Lint with integration tests enabled: zero issues. `make sqlc-check` and `git diff --check` passed.

The broader `TestAgentExecutorKeepsPromptPrefixStableAcrossTurns` integration test was attempted but failed during fixture setup because local object storage returned `403 AccessDenied` while uploading a skill archive.


cmsparks (approved):
lgtm!

PR #551, opened 2026-09-17
refactor: store internal config IDs and expose saved compiled configs

asherdale:
- Store compiled resource references and machine secret-environment references as internal UUIDs, converting to public IDs at API boundaries.
- Expose saved compiled configurations on detail responses while keeping profile lists and organization overviews lean.
- Show read-only compiled JSON for subagents and beneath agent/profile YAML editors.
- Migrate existing stored references without modifying authored YAML/JSON, and drop the obsolete `agent_configs.definition` column.


migrations/000038_explicit_default_tools_test.go
```diff
@@ -16,7 +16,7 @@ func TestExplicitDefaultToolsMigration(t *testing.T) {
 	require.NoError(t, err)
 	opts := agentconfig.CompileOptions{
 		ResolveSkillID: func(id string) (agentconfig.SkillResolution, error) {
-			return agentconfig.SkillResolution{PublicID: id, Name: "test-skill"}, nil
+			return agentconfig.SkillResolution{ID: uuid.Must(publicid.Decode(publicid.KindSkill, id)), Name: "test-skill"}, nil
```
ksarangmath, line 19: nit: we can probably completely delete this test file, once the migration is done, we don't have use for this test file (in general I don't think we need test files for migrations, esp after the migration has already occured)

internal/agentconfig/compiler.go
```diff
@@ -9,13 +9,14 @@ import (
...
 
+	"github.com/google/uuid"
 	"github.com/omnara-ai/omnara/internal/publicid"
 	"github.com/omnara-ai/omnara/internal/secrets"
 	"github.com/omnara-ai/omnara/internal/toolcatalog"
 	"github.com/omnara-ai/omnara/internal/toolpermission"
 )
 
-const CompilerVersion = ""
+const CompilerVersion = "1"
```
ksarangmath, line 19: nit: Can we remove compiler_version as part of this cleanup? We support one compiled-config format, and this migration upgrades all existing configs to that format. The version is carried through storage, runtime calls, and the API, but its runtime behavior is just to reject anything other than the current value. Having migrations own the format transition seems sufficient for the current design. We can introduce explicit format versioning when we have a concrete need for multiple formats.

internal/httpapi/compiled_config.go
```diff
@@ -0,0 +1,122 @@
...
+	"bytes"
+	"encoding/json"
+	"fmt"
+
+	"github.com/google/uuid"
+	"github.com/omnara-ai/omnara/internal/publicid"
+)
+
+func publicCompiledDefinition(raw json.RawMessage) (json.RawMessage, error) {
+	var root map[string]any
```
ksarangmath, line 13: Can we make this an explicit API type, and have the id references use a type from thez publicid package?

docs/api-reference/openapi.yaml
```diff
@@ -8944,6 +8944,12 @@ components:
         source_format:
           type: string
           enum: [yaml, json]
+        compiled_definition:
```
ksarangmath, line 8947: nit: should use separate API response types for summaries and full configs? The summary type could omit compiled_definition, and the full config type could require it. Currently they share one type, so the field is optional for endpoints that always return it.

internal/agentconfig/subagents.go
```diff
@@ -57,7 +58,7 @@ type AgentConfigSubagentInstructionSource struct {
 
 type SubagentCompiled struct {
 	Type                    string                 `json:"type"`
-	ProfileID               string                 `json:"profile_id,omitempty"`
+	ProfileID               uuid.UUID              `json:"profile_id,omitzero"`
 	Description             string                 `json:"description,omitempty"`
 	Model                   *SubagentModelCompiled `json:"model,omitempty"`
```
ksarangmath, line 63: This comment predates this PR, but I think we should make subagent model selection consistent with regular agent configs as part of this cleanup

When selecting a model, can we require both provider_config and name, resolve them during compilation, and save configured_model_id? If neither is provided, inherit the parent’s model for self subagents or the profile’s current model for profile subagents.

This would require a migration that changes source and configs, I think the tricky part would be to migrate sources that only defined one of name or provider_config, but hopefully those sources don't exist yet, which you could check

ksarangmath (commented):
LGTM, just left comments on some things I think we should add in this PR as cleanup not caused by this PR, but  it will make things cleaner going forward if we just do it in this migration

internal/httpapi/ids.go
```diff
@@ -1,12 +1,35 @@
...
+	"encoding/json"
 	"fmt"
 
 	"github.com/google/uuid"
+	"github.com/omnara-ai/omnara/internal/httpapi/apierror"
 	"github.com/omnara-ai/omnara/internal/publicid"
+	"github.com/omnara-ai/omnara/internal/storage/storeerr"
 )
 
+func secretIDsFromPointer[T any](value *T) (json.RawMessage, error) {
```
ksarangmath, line 13: nit: can we use explicit secret-env map types in these helpers instead of any?

migrations/000038_explicit_default_tools_test.go
asherdale: removed, also removed the migration-specific tests for this new migration after validating against a local production-data copy. they’re available in commit `da800e7e9` if needed

internal/agentconfig/compiler.go
asherdale: removed `compiler_version` from application code and the API

internal/httpapi/compiled_config.go
asherdale: done - `compiled_definition` now uses an explicit `CompiledAgentConfig` API type, with ID references typed as `publicid.ID`

docs/api-reference/openapi.yaml
asherdale: good call, split config/profile responses into summary and full types. profile lists and the org overview omit `compiled_definition`, full responses require it

internal/agentconfig/subagents.go
asherdale: selecting a subagent model now requires both `provider_config` and `name`, resolves during compilation, and stores `configured_model_id`. omitted selections inherit the parent’s model for `self` or the profile’s current model for `profile`. removed the spawn-time name resolver

extended migration 40 to resolve existing selections and fill the missing source field for legacy partial selections. audited all 2,934 production configs: no explicit or partial subagent model selections. kept partial-selection handling for other deployments and configs created before rollout. migrations 40/41 passed against the local production-data copy

migrations/000042_internal_config_ids.go
```diff
@@ -0,0 +1,510 @@
...
+		}
+		model, ok := subagent["model"].(map[string]any)
+		if !ok {
+			return fmt.Errorf("subagent %q model must be an object", key)
+		}
+		provider, hasProvider := model["provider_config"]
+		name, hasName := model["name"]
+		if !hasProvider && !hasName {
+			continue
+		}
```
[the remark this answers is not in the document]
asherdale: these fields were persisted in the old `SubagentModelCompiled`: `compileSubagents` explicitly copied `provider_config` and `name` into it. migration 40 reads the stored JSON directly, not through the new `ModelCompiled`, so it can resolve those selections and repair partial sources. the new type only applies after conversion

api/openapi/openapi.yaml
```diff
@@ -8927,7 +8927,266 @@ components:
...
+          $ref: '#/components/schemas/ConfiguredModelID'
+          x-go-type: publicid.ID
+          x-go-type-import:
+            path: github.com/omnara-ai/omnara/internal/publicid
+          x-go-type-skip-optional-pointer: true
+        context_window_tokens:
+          type: integer
+        default_max_output_tokens:
+          type: integer
+        cache_retention:
```
ksarangmath, line 9009: nit: can we use enums for these fields that have small sets of values? like cache_retention, tool type, mcp auth type, subagent type. ModelCacheRetention is already an existing enum we could user here

api/openapi/openapi.yaml
```diff
@@ -8927,7 +8927,266 @@ components:
...
+          x-go-type-import:
+            path: github.com/omnara-ai/omnara/internal/toolpermission
+        deferred:
+          type: boolean
+    CompiledSkill:
+      type: object
+      additionalProperties: false
+      required: [public_id]
+      properties:
+        public_id:
```
ksarangmath: nit: we can name this just id, not public id (same for any other field named public id, since to external users, the id is the public id)

api/openapi/openapi.yaml
```diff
@@ -8927,7 +8927,266 @@ components:
...
+          $ref: '#/components/schemas/SkillID'
+          x-go-type: publicid.ID
+          x-go-type-import:
+            path: github.com/omnara-ai/omnara/internal/publicid
+    CompiledSubagent:
+      type: object
+      additionalProperties: false
+      required: [type]
```
ksarangmath, line 9222: very nit (i'm fine if you don't do this): we could discriminate on the type and variants, one for self, one for profile, and profile requires profile_id, self doesn't.

api/openapi/openapi.yaml
```diff
@@ -8927,7 +8927,266 @@ components:
...
+          x-go-type-skip-optional-pointer: true
+        tools:
+          type: object
+          x-go-type-skip-optional-pointer: true
+          additionalProperties:
+            $ref: '#/components/schemas/CompiledMCPTool'
+    CompiledMCPAuth:
+      type: object
+      additionalProperties: false
+      required: [type, secret_id]
```
ksarangmath, line 9142: similar to above comment, very nit, but sigv4 could be expressed explicitly with needing service and region. basically we could be more explicit/detailed in the openapi schema, but not a big deal for these specific cases. For bigger things, we should be explicit (like what this change is doing, I haven't found anything that really needs to be changed)

ksarangmath (approved):
LGTM!

api/openapi/openapi.yaml
asherdale: reused `ModelCacheRetention` and added enums for tool, auth, and subagent types

api/openapi/openapi.yaml
asherdale: done, renamed it to `id`. no `public_id` fields remain in the OpenAPI spec

api/openapi/openapi.yaml
asherdale: makes sense, split into `self` and `profile` variants discriminated by `type`

api/openapi/openapi.yaml
asherdale: split auth into `bearer`, `oauth`, and `sigv4` variants

internal/agentconfig/subagents.go
```diff
@@ -187,16 +137,24 @@ func compileSubagents(
...
+			if entry.Model.ProviderConfig != "" && opts.ResolveModelSelection != nil {
+				resolved, err := opts.ResolveModelSelection(entry.Model.ProviderConfig, entry.Model.Name)
+				if err != nil {
+					return nil, issueAt(pointer, issueOr("/model", err))
+				}
+				if resolved.ConfiguredModelID == uuid.Nil {
+					return nil, issuef(jsonPointer("subagents", key, "model"), "resolver returned an empty model id")
+				}
+				out.Model.ConfiguredModelID = resolved.ConfiguredModelID
+			}
```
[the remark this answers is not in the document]
asherdale: name-only never reaches this: the source schema's `DependentRequired` (name ↔ provider_config) rejects it in `parseSource` before compile, for api and cli alike

PR #552, opened 2026-09-18
chore: Worktree mcp tool annotations and oidc userinfo

cmsparks:
Make some auth and tool annotation improvements so tool scanning and auth works for claude connectors and chatgpt plugins

Claude requires:
* clearer tool types
* title annotations

chatgpt requires:
* additional oauth support (specifically things listed here https://developers.openai.com/plugins/build/auth#custom-auth-with-oauth-21 )
* oidc user info for certain enterprise support (not required but recommended)


internal/storage/queries/identity.sql
```diff
@@ -1424,21 +1427,22 @@ WITH presented AS (
     )
 ), rotated AS (
   UPDATE oauth_access_tokens token
-  SET token_hash = sqlc.arg(token_hash),
+  SET scope = CASE WHEN sqlc.arg(scope)::text = '' THEN token.scope ELSE sqlc.arg(scope)::text END,
+      token_hash = sqlc.arg(token_hash),
```
ishaansehgal99: This scope field is currently serving two purposes: the permissions the user originally approved, and the permissions on the current access token. Updating it here means requesting a temporarily narrower access token permanently reduces the stored grant.

Example: the user approves openid email, then the client refreshes with openid. That access token should exclude email, but a subsequent refresh requesting openid email should still succeed because the user already approved it.

internal/httpapi/apimcp/server.go
```diff
@@ -141,20 +142,15 @@ func NewHandler(server *mcp.Server) http.Handler {
...
 }
 
-func annotationsFor(method string, destructive bool) *mcp.ToolAnnotations {
+func annotationsFor(tool Tool, method string) *mcp.ToolAnnotations {
 	readOnly := method == http.MethodGet
-	idempotent := readOnly || method == http.MethodPut || method == http.MethodDelete
-	annotations := &mcp.ToolAnnotations{
-		ReadOnlyHint:   readOnly,
-		IdempotentHint: idempotent,
+	destructive := !readOnly && (tool.Destructive || method == http.MethodDelete)
```
ishaansehgal99: as a heads up this now marks destructiveHint: false for every non-DELETE operation. this includes update operations like crons_update, pools_update, and agents_update

internal/storage/identitystore/oidc_signing_key.go
```diff
@@ -0,0 +1,74 @@
...
+		secrets.Payload{secrets.KeyValue: base64.StdEncoding.EncodeToString(x509.MarshalPKCS1PrivateKey(key))},
+		oidcSigningKeyAAD())
+	if err != nil {
+		return fmt.Errorf("encrypt OIDC signing key: %w", err)
+	}
+	payload, err := json.Marshal(encrypted)
+	if err != nil {
+		return fmt.Errorf("encode OIDC signing key: %w", err)
+	}
+	if err := s.q.CreateOIDCSigningKey(ctx, dbsqlc.CreateOIDCSigningKeyParams{EncryptedPrivateKey: payload}); err != nil {
```
ishaansehgal99, line 63: Non-blocking follow-up: the existing encryption-key rotation operation only updates secret_versions, so it will miss this new signing-key record. Removing the old encryption key afterward would break connector authorization and token refresh

ishaansehgal99 (approved):
LGTM! nits

internal/storage/queries/identity.sql
cmsparks: Ah yeah good catch, fixing

internal/httpapi/apimcp/server.go
cmsparks: Yeah I'll update this, I understood destructiveHint differently before

internal/storage/identitystore/oidc_signing_key.go
cmsparks: sg, will follow up

PR #553, opened 2026-09-18
feat: Implement builder UI improvements

cmsparks:


frontend/apps/web/src/components/ui/dialog.tsx
```diff
@@ -25,10 +25,7 @@ function DialogOverlay({ className, ...props }: ComponentProps<typeof DialogPrim
   return (
     <DialogPrimitive.Overlay
       data-slot="dialog-overlay"
-      className={cn(
-        'data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 bg-overlay fixed inset-0 z-50',
-        className,
-      )}
```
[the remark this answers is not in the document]
cmsparks: this is expected

frontend/apps/web/src/components/agents/AgentConfigMachineSourceComboboxes.tsx
```diff
@@ -251,3 +260,87 @@ export function MachineSourceCombobox({
...
   )
 }
+
+export interface ResolvedPoolGrant {
+  machine_pool: MachinePoolSummary
+  grant: ProjectMachinePoolGrant
+}
+
+export function PoolGrantSummary({ machine_pool: pool, grant }: ResolvedPoolGrant) {
+  const poolQuery = useMachinePool(pool.org_id, pool.id)
```
asherdale, line 270: this endpoint requires org management permission, so project developers get a 403 and see inherited defaults as "Unset". could we distinguish unavailable values from unset ones?

frontend/apps/web/src/components/agents/mcpAvailability.ts
```diff
@@ -0,0 +1,91 @@
...
+    case 'disabled':
+      return { defaultEnabled: false, deferred: undefined }
+  }
+}
+
+export function mcpToolAvailability(
+  tool: BasicMcpTool,
+  serverDeferred: boolean,
+): McpToolAvailability {
+  if (tool.enabled === false) return 'disabled'
```
asherdale, line 71: opening YAML with `default_enabled: false` and a tool override containing only `deferred: false` shows "Enabled" in the builder, but the tool stays disabled. could this account for the server default?

frontend/apps/web/src/components/agents/AgentConfigSkillsField.tsx
```diff
@@ -157,10 +169,30 @@ export function AgentConfigSkillsField({
...
       ) : null}
+      {createFor !== null && (
+        <CreateSkillDialog
+          open
+          onOpenChange={(open) => {
+            if (!open) setCreateFor(null)
+          }}
+          orgId={orgId}
+          owner={{ kind: 'project', project_id: projectId }}
+          onCreated={(skill) => {
```
asherdale, line 196: using add skill → create skill to re-upload an attached skill appends its existing id again, which the backend rejects as a duplicate. could we avoid adding it twice?

frontend/apps/web/src/components/machines/MachineOverrideFields.tsx
```diff
@@ -167,128 +168,178 @@ export function CombinedEnvOverlayEditor({
...
             onChange({ envRows: [...envRows, newEnvOverlayRow()], secretEnvRows })
           }}
         >
+          <PlusIcon />
           Add variable
         </Button>
       </div>
+      <div className="overflow-hidden rounded-xl border">
+        <div className="after:border-border/50 after:bg-muted/50 after:shadow-xs relative isolate after:pointer-events-none after:absolute after:-inset-x-px after:-top-px after:-z-10 after:h-[calc(2.25rem+2px)] after:rounded-xl after:border">
+          <Table className="table-fixed">
```
asherdale: this replaces the mobile two-row layout with a fixed table, squeezing the env inputs to almost nothing. the MCP overrides table also hides tool names at phone width. could we make both responsive?

frontend/apps/web/src/components/agents/AgentEventLog.tsx
```diff
@@ -117,9 +117,9 @@ function SendInputGuide({
       <CodeTabsBlock
         label="How to send input to this agent"
         tabs={[
-          { value: 'cli', label: 'CLI', content: commands.cli },
           { value: 'sdk', label: 'TypeScript SDK', content: commands.sdk },
           { value: 'curl', label: 'cURL', content: commands.curl },
+          { value: 'cli', label: 'CLI', content: commands.cli },
```
asherdale: should we move (and slightly change the wording of) the helpful 'copy and run' text from the CLI snippet to the SDK snippet, now that the SDK snippet is the first one shown?

Image

Image

asherdale: i think kartik had also mentioned adding some new UI encouraging users to build on top of the API. there are a few ways we could do this obviously - for example one way could be to extend the existing sentence below the snippets to something like this: 'build your own custom frontend on our API, or send a message via the Chat'

frontend/apps/web/src/components/agents/AgentConfigMachineSourceOverrides.tsx
```diff
@@ -1,109 +1,128 @@
...
-              idPrefix={source.id}
-              pool={{ provider: source.provider, management_kind: source.managementKind }}
-              values={source.providerOptions}
-              onChange={(providerOptions) => {
-                onChange({ providerOptions })
+    <FieldGroup>
+      {source.kind === 'pool' && (
+        <div className="grid gap-4 sm:grid-cols-2">
+          <Field>
+            <FieldLabel htmlFor={`${source.id}-initial`}>Initial machines</FieldLabel>
```
asherdale: Image

could we move the row with initial machines / max machines above the working directory row? IMO it's the second-most important row (besides the machine memory/cpu)

asherdale: i also noticed that these advanced machine source options appear expanded as soon as you add a new source, was that intentional? i wonder if it's a friendlier experience if they stay hidden/collapsed until you select an actual pool/BYO from the dropdown (the same behavior when adding a subagent), or if they just stay hidden/collapsed until intentionally opened

asherdale (approved):
lgtm 💪 comments are all small issues/nits, so should just be straightforward fixes but lmk if you want another look!

frontend/apps/web/src/components/agents/AgentConfigMachineSourceOverrides.tsx
cmsparks: Yeah the intention is that when you select it it autoexpands once. IMO it's most likely when you're configuring it for the first time

frontend/apps/web/src/components/agents/AgentEventLog.tsx
cmsparks: yeah i'll update this copy a bit to encourage building on the API

frontend/apps/web/src/components/machines/MachineOverrideFields.tsx
cmsparks: adding responsiveness

PR #554, opened 2026-09-18
fix: avoid locking unchanged actor rows on identity upsert

asherdale:
- `UpsertActorIdentity` uses `ON CONFLICT ... DO UPDATE ... WHERE <display name changed>`. Postgres locks the conflicting row before evaluating that `WHERE`, so even a no-op upsert takes a row lock and holds it until commit.
- Every request from one principal resolves the same actor row, so concurrent archives, launches, and inputs from a single API key or user serialized on it. Archive takes the lock early and then does the whole archive tree, and launch waits on it while holding the per-project agent-creation advisory lock, so creates stalled too. This is what caused the `omnara-prod-postgres-lock-wait-burst` alarms.
- `upsertActorIdentityTx` now reads the actor first and returns it directly when the upsert would be a no-op: the actor exists and the incoming display name is blank or already matches. That fast path takes no row lock.
- Missing actors and real display-name changes still go through the existing upsert, and the `ErrNoRows` re-read still covers concurrent creation of the same identity.
- The fast-path check only needs `display_name`: the upsert never writes the identity columns, `id`, `created_at`, or `metadata`, and it only bumps `updated_at` when the display name changes.
- `PutActor` (external actors) has the same pattern but isn't in the hot path here, so it's unchanged.


cmsparks (approved):
lgtm

PR #556, opened 2026-09-18
feat: add configurable event webhooks

asherdale:
- Add `event_webhook` with a public HTTPS URL, optional Standard Webhooks signing, and an explicit event-type selection.
- Configure webhooks in the builder’s collapsed Advanced section, defaulting to tool-call updates.
- Queue selected events transactionally and retry failed deliveries for up to ten minutes, with per-organization concurrency limits and claim recovery.
- Reuse SSE data payloads and provide webhook examples, OpenAPI definitions, and generated SDK types and Zod validators.


frontend/apps/web/src/components/agents/agentConfigEventWebhook.ts
```diff
@@ -0,0 +1,35 @@
...
+export function eventWebhookUrlError(value: string): string | undefined {
+  if (value.trim() === '') return undefined
+  try {
+    const url = new URL(value.trim())
+    if (url.protocol !== 'https:') return 'Use an HTTPS URL.'
+  } catch {
+    return 'Enter a valid HTTPS URL.'
+  }
+  return undefined
+}
```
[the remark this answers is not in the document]
asherdale: consistent with our other URL fields. backend validation errors already show in the builder; no need to duplicate those checks

internal/eventwebhook/sender.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+	if err != nil && delivery.Retryable {
+		delay := retryDelay(delivery.AttemptCount)
+		err = s.store.RetryEventWebhookDelivery(updateCtx, delivery.ID, delivery.ClaimToken, delay)
+	} else {
+		err = s.store.CompleteEventWebhookDelivery(updateCtx, delivery.ID, delivery.ClaimToken)
+	}
+	if err != nil {
+		s.log.Warn("update event webhook delivery", "delivery_id", delivery.ID, "error", err)
+	}
```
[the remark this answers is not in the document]
asherdale: the token check already protects newer claims. checking affected rows would only add logging, not prevent duplicates

docs/agents/configuration.mdx
```diff
@@ -154,6 +154,23 @@ The full example above defines `create_ticket` as a custom tool. When the agent
...
+## Event webhook
+
+`event_webhook` sends this agent's timeline events and tool-call updates to a public HTTPS endpoint:
+
+```yaml
+event_webhook:
+  url: https://example.com/events
+```
+
+Set `event_webhook.events` to an event-type allowlist, such as `[tool_call_update, model_output]`. Supported event types are `agent_input`, `model_output`, `tool_result`, `context_checkpoint`, and `tool_call_update`. Omit the list to send all supported event types; an explicit list must contain at least one type. Changes apply to newly queued events.
```
cmsparks: > Omit the list to send all supported event types;

We shouldn't do this as the default IMO, this will cause problems like GH with scaling later. Github defaults to sending everything, which increases the load on our end. We can have an explicit option (maybe "*"), but it should be opt in to send every event rather than opt out.

cmsparks: The format for sending events should be better documented here. If we're following Standard Webhooks, we need to document the full payloads for every event: https://github.com/standard-webhooks/standard-webhooks/blob/main/spec/standard-webhooks.md#payload

Github does this pretty comprehensively: https://github.com/standard-webhooks/standard-webhooks/blob/main/spec/standard-webhooks.md#payload

It's also nice if it'd be somehow autogenerated (maybe from our openapi spec? not sure the best way to do this). I think it's best for us to define the payloads as types in the openapi specification (even though they're not directly used by any route). Then they can get converted into zod schemas, and users can use the zod schemas to parse incoming webhooks on their own servers

docs/agents/configuration.mdx
```diff
@@ -154,6 +154,23 @@ The full example above defines `create_ticket` as a custom tool. When the agent
...
+
+Each POST contains `event` and `data` matching the [event stream](/events/streaming), excluding token previews, heartbeats, and stream errors. Each agent sends its own events. Self-subagents inherit the webhook configuration and send events with their own agent IDs; other subagents use their profile’s configuration.
+
+Deliveries are queued for up to ten minutes, with a five-second request timeout. Failed `tool_call_update` deliveries are retried within that window for custom tools entering `ready`, any tool entering `awaiting_permission`, and the built-in `ask_question` entering `running`. All other events are best effort, with no scheduled retries after failure. Return `2xx` to acknowledge receipt. Events may arrive more than once or out of order; deduplicate custom tool execution by tool-call ID.
```
cmsparks: Why are only failed `tool_call_update`s retried? We should have the same retry behavior for all delivery types.

cmsparks (commented):
leaving these (mainly documentation) comments initially, going to continue reviewing further

docs/agents/configuration.mdx
asherdale: good call - `events` is now required and nonempty, and the builder defaults to `tool_call_update` only

asherdale: added examples for every event, OpenAPI payload types, and generated Zod schemas. kept `{event, data}` to match the event stream

docs/agents/configuration.mdx
asherdale: def agree, originally when the default was 'All Events' I was concerned about causing the 'thundering herd' problem for webhook receivers that had minor/transient outages with significant agent activity, but now that it is explicitly opt-in for everything based on your earlier comment, i've changed it so all selected event types now use the same retry policy

docs/events/webhooks.mdx
```diff
@@ -0,0 +1,155 @@
...
+description: "Event webhook bodies and generated SDK schemas"
+---
+
+Configure an [event webhook](/agents/configuration#event-webhook) to receive JSON POST requests with `event` and `data`. The event name determines the data schema. These are the same data objects used by the [event stream](/events/streaming).
+
+Every request includes `Webhook-Id` (stable across retries) and `Webhook-Timestamp` (Unix seconds for that attempt). Signed requests also include `Webhook-Signature`. Verify the signature against the raw request body before parsing it, using [Standard Webhooks](https://github.com/standard-webhooks/standard-webhooks/blob/main/spec/standard-webhooks.md).
+
+## Parse with the SDK
+
+`EventWebhookPayload` and `zEventWebhookPayload` are generated from the [OpenAPI specification](/api-reference/openapi.yaml). The schema checks that `event` matches the shape of `data`.
```
cmsparks:
```suggestion
The SDK includes types `EventWebhookPayload` and `zEventWebhookPayload` to validate webhook events on your webhook receiver endpoint.
```

docs/events/webhooks.mdx
```diff
@@ -0,0 +1,155 @@
+---
+title: "Webhook payloads"
+description: "Event webhook bodies and generated SDK schemas"
+---
+
+Configure an [event webhook](/agents/configuration#event-webhook) to receive JSON POST requests with `event` and `data`. The event name determines the data schema. These are the same data objects used by the [event stream](/events/streaming).
+
+Every request includes `Webhook-Id` (stable across retries) and `Webhook-Timestamp` (Unix seconds for that attempt). Signed requests also include `Webhook-Signature`. Verify the signature against the raw request body before parsing it, using [Standard Webhooks](https://github.com/standard-webhooks/standard-webhooks/blob/main/spec/standard-webhooks.md).
```
cmsparks, line 8: We should have a code snippet which shows creating a simple server which receives, parses, and validates a webhook (with the signature). Important part is a code sample for how to verify the signature, or else people will create insecure apps

cmsparks: We should also add an example of custom tools + webhooks in the examples/ folder IMO, because this is a slightly more complex usecase. It probably will only end up being like 4 files, should be pretty simple

internal/eventwebhook/sender.go
```diff
@@ -0,0 +1,220 @@
...
+				}
+			}
+			claimCtx, cancel := context.WithTimeout(ctx, storeTimeout)
+			delivery, err := s.store.ClaimEventWebhookDelivery(claimCtx, excluded)
+			cancel()
+			if err == nil {
+				inFlight++
+				activeByOrg[delivery.OrgID]++
+				deliveries.Go(func() {
+					defer func() { completed <- delivery.OrgID }()
```
cmsparks: this very odd code structure, why are we not just doing completed <- delivery.OrgID

this just defers the function to run on the next line

cmsparks: the only difference is if deliver panics, but:
1) that should never happen and is a sign of a more significant issue
2) if there really is a chance of a failure, we should instead be returning well formed errors

internal/eventwebhook/sender.go
```diff
@@ -0,0 +1,220 @@
...
+		client: outboundhttp.NewPublicClient(outboundhttp.PublicClientOptions{Timeout: attemptTimeout})}
+}
+
+func (s *Sender) Run(ctx context.Context) {
+	defer s.client.CloseIdleConnections()
+	perOrgLimit := max(1, s.maxInFlight/perOrgConcurrencyDivisor)
+	activeByOrg := make(map[uuid.UUID]int)
+	completed := make(chan uuid.UUID, s.maxInFlight)
+	inFlight := 0
+	var deliveries sync.WaitGroup
```
cmsparks, line 67: We're triple counting here.

activeByOrg holds the count of active
inFlight holds the count of active
and deliveries holds the count of active

only one of these is actually concurrency safe (the waitgroup)

For the per org limits, we should move that into the query rather than the application code (esp because it being in the app code is potentially racey).

internal/eventwebhook/sender.go
```diff
@@ -0,0 +1,220 @@
...
+			if !errors.Is(err, storeerr.ErrNotFound) && ctx.Err() == nil {
+				s.log.Warn("claim event webhook", "error", err)
+			}
+		}
+		var timer *time.Timer
+		var idle <-chan time.Time
+		if inFlight < s.maxInFlight {
+			timer = time.NewTimer(idleInterval)
+			idle = timer.C
+		}
```
cmsparks, line 89: we can do a time.After instead here

internal/eventwebhook/sender.go
cmsparks: Here's a sample of rewriting run with some taking into account some of my comments in this file:

```go
func (s *Sender) Run(ctx context.Context) {
      defer s.client.CloseIdleConnections()
      slots := make(chan struct{}, s.maxInFlight)
      defer func() {
              for range s.maxInFlight {
                      slots <- struct{}{}
              }
      }()
      ticker := time.NewTicker(idleInterval)
      defer ticker.Stop()
      for {
              select {
              case <-ctx.Done():
                      return
              case slots <- struct{}{}:
              }
              claimCtx, cancel := context.WithTimeout(ctx, storeTimeout)
              delivery, err := s.store.ClaimEventWebhookDelivery(claimCtx, s.perOrgLimit)
              cancel()
              if err == nil {
                      go func() {
                              s.deliver(ctx, delivery)
                              <-slots
                      }()
                      continue
              }
              <-slots
              if !errors.Is(err, storeerr.ErrNotFound) && ctx.Err() == nil {
                      s.log.Warn("claim event webhook", "error", err)
              }
              select {
              case <-ctx.Done():
                      return
              case <-ticker.C:
              }
      }
}
```

A lot of the state doesn't actually need to be handled by this function and can be offloaded to go's sync primitives or to the DB I think

cmd/maintenance/main.go
```diff
@@ -387,6 +389,12 @@ func runCoreMaintenanceTick(
 			executionstore.ProcessToolMachineUnreachableGrace,
 		)
 	expireProcessToolsOutcome := completedMaintenanceOutcome(ctx, expireProcessToolsErr)
+	webhookCleanupCtx, cancelWebhookCleanup := context.WithTimeout(ctx, eventWebhookCleanupTimeout)
```
cmsparks: We're adding a ton of random cleanup to this "runCoreMaintenanceTick" function. It probably makes sense to wrap this now I think. Maybe some dedicated store function which wraps all of these, and then does helpers. We probably also don't need to chain them all in a row, they can probably be done in parallel / through goroutines (i.e. deleting webhooks doesn't need to be blocked by expiring processes)

cmsparks: can also be a follow up if we want

internal/eventwebhook/sender.go
```diff
@@ -0,0 +1,220 @@
...
+	client      *http.Client
+	log         *slog.Logger
+}
+
+func New(store Store, log *slog.Logger, maxInFlight int) *Sender {
+	return &Sender{store: store, log: log, maxInFlight: maxInFlight,
+		client: outboundhttp.NewPublicClient(outboundhttp.PublicClientOptions{Timeout: attemptTimeout})}
+}
+
+func (s *Sender) Run(ctx context.Context) {
```
cmsparks, line 63: We should do a structured event log for every loop, like we do with workers. This is more important because we delete queued delivery rows upon failure. We should at least track those failures in our logs I think

cmsparks: Can probably do something like the following:

- logent.EventWebhookDelivery(ctx, delivery, target) creating a event_webhook.delivery event per attempt, attaching org/project/agent ids, delivery id, event kind or tool call id, attempt count, and target URL host. Call event.Done at the end of deliver.
- logent.EventWebhookDeliveryResult(ctx, status, err, retryAt) attaching HTTP status, error, and next attempt. Success stays info or debug, failure with a retry scheduled stays warn, and the final attempt that lands past the window logs at error with a gave_up: true field. That covers the silent-drop finding.
- logent.EventWebhookStoreFailed(ctx, op, err) for claim and complete/retry store errors, instead of the bare warns.

internal/eventwebhook/sender.go
```diff
@@ -0,0 +1,220 @@
...
+				inFlight++
+				activeByOrg[delivery.OrgID]++
+				deliveries.Go(func() {
+					defer func() { completed <- delivery.OrgID }()
+					s.deliver(ctx, delivery)
+				})
+				continue
+			}
+			if !errors.Is(err, storeerr.ErrNotFound) && ctx.Err() == nil {
+				s.log.Warn("claim event webhook", "error", err)
```
cmsparks: see above, should be using our structured logger instead of adhoc logs

internal/eventwebhook/sender.go
```diff
@@ -0,0 +1,220 @@
...
+func (s *Sender) deliver(ctx context.Context, delivery executionstore.EventWebhookDelivery) {
+	attemptCtx, cancel := context.WithTimeout(ctx, attemptTimeout)
+	err := s.send(attemptCtx, delivery)
+	cancel()
+	if ctx.Err() != nil {
+		return
+	}
+	updateCtx, updateCancel := context.WithTimeout(ctx, storeTimeout)
+	defer updateCancel()
+	if err != nil {
```
cmsparks, line 111: this error handling is odd:
1. we're checking the same error twice in a row on 122 and 125
2. we're reassigning error in 127 and 129, we should have separate error handling for retry and complete failures
3. we should be using our structured logger instead

docs/events/webhooks.mdx
asherdale: good call, added a server snippet that verifies the raw body before schema validation, plus `examples/custom-tool-webhook` showing custom-tool execution, duplicate handling, and result submission

internal/eventwebhook/sender.go
asherdale: removed the defer, completion is sent directly after `deliver` returns

cmd/maintenance/main.go
asherdale: handled here: extracted `runCoreMaintenance` into `cmd/maintenance/core.go` and made the five cleanup tasks run concurrently. kept orchestration in maintenance, with the individual operations in their stores

internal/eventwebhook/sender.go
asherdale: added a structured event per attempt with the requested fields, warn-level retries, and error-level `gave_up` on the final failed attempt. claim, complete, and retry store failures use `EventWebhookStoreFailed`

internal/eventwebhook/sender.go
asherdale: yep you right, removed the duplicate check, separated retry and completion errors, and routed the failures through the structured logger

internal/eventwebhook/sender.go
asherdale: good call, moved the per-org limit into the query and removed `activeByOrg`. the remaining counter is only touched by `Run`, so it’s safe - the channel gives us immediate refill and the `WaitGroup` handles shutdown. the shared limit is a soft cap under simultaneous claims

cmd/maintenance/core.go
```diff
@@ -0,0 +1,77 @@
+package main
```
cmsparks: nit: typically, cmd packages only have the entrypoint (main.go) inside of them, can we move these files to a separate package?

examples/custom-tool-webhook/server.ts
```diff
@@ -0,0 +1,135 @@
+import { createServer } from 'node:http'
```
cmsparks: mind if we use hono (or express, but hono is what most ppl use nowadays) instead of raw http? should be much much shorter

examples/custom-tool-webhook/server.ts
```diff
@@ -0,0 +1,135 @@
...
+      if (payload.event === 'tool_call_update' && payload.data.state === 'ready') {
+        const { agent_id: agentID, tool_call_id: toolCallID } = payload.data
+        if (!agentID) {
+          response.writeHead(400).end('missing agent_id')
+          return
+        }
+        let work = inFlight.get(toolCallID)
+        if (!work) {
+          work = execute(agentID, toolCallID)
+          inFlight.set(toolCallID, work)
```
cmsparks: nit: i don't think need to track inflight because the request to submitToolCallResult will just fail (and the other one will 200), and that should be fine.

aside: our examples probably don't need to be 1000% error proof, we probably don't need excesssive tests + error handling. tbh it's probably better if they're simpler + easier to read than 100% correct, users can add their app specific error handling later

cmsparks (commented):
lgtm, mostly just nits

cmsparks (approved):

examples/custom-tool-webhook/.gitignore
```diff
@@ -0,0 +1,3 @@
+node_modules/
+.env
+pnpm-lock.yaml
```
cmsparks: probably don't need this gitignore, should already be handled by root gitignore

asherdale: def agree, but wasn't handled previously. updated the root gitignore to handle these files, which then meant i could delete the gitignore from all of the other examples too

PR #560, opened 2026-09-19
refactor: add structured database error details to request logs

asherdale:
- Add structured DB error kinds and SQLSTATE codes, and prioritize failed queries so their names remain visible within the request log limit.
- Record cancellation sources and request cancellation state to distinguish database timeouts from SSE and socket teardown.
- Replace raw DB error messages with safe classifications across request, auth, and worker logs to avoid exposing SQL arguments or customer content.
- Resolve [OMN2-344](https://linear.app/omnara/issue/OMN2-344/add-structured-database-error-details-to-wide-request-logs).


internal/log/event.go
```diff
@@ -182,9 +190,16 @@ func (e *Event) Done(ctx context.Context) {
 	attrs := append([]slog.Attr(nil), e.attrs...)
 	attrs = append(attrs, slog.Duration("event.duration", time.Since(e.started)))
 	if e.err != nil {
-		attrs = append(attrs, slog.String("error.message", e.err.Error()))
+		attrs = append(attrs, slog.Any("error.message", e.err))
 	}
 	attrs = dedupAttrs(attrs)
+	for i, attr := range attrs {
+		if err, ok := attr.Value.Any().(error); ok {
+			if text, ok := e.databaseErrorText(err); ok {
```
ksarangmath: Could we make DB errors safe to print when they leave storage, rather than determine that from this event’s query history? The same error can be sanitized here but printed raw in a child event or by a helper that already converted it to a string. A safe description carried by the error itself, while preserving the underlying cause for error checks, would make this independent of tracing and where the error gets logged.

internal/httpapi/daemon_socket.go
```diff
@@ -192,8 +195,7 @@ func (s *daemonSocket) run(ctx context.Context) {
 			}
 		}
 	}()
-	s.readLoop(runCtx)
-	cancel()
+	cancel(s.cancellationCause(s.readLoop(runCtx)))
```
ksarangmath: nit, this was preexisting, but can we make the socket own its background drain work? We wait for the writer here, but not the drain, so the request log can finish before a canceled DB query reports its error, dropping these diagnostics. Redis-triggered drains also use a different context and can report socket_closed after a socket failure.

internal/log/tracing.go
```diff
@@ -87,27 +104,44 @@ func (e *Event) flushDBQueries() {
...
+		emitted++
 		out[prefix+"name"] = record.Name
 		out[prefix+"start_time_ms"] = relativeMilliseconds(record.Start, e.started)
-		out[prefix+"duration_ms"] = durationMs
+		out[prefix+"duration_ms"] = record.Duration.Milliseconds()
 		out[prefix+"rows"] = record.Rows
-		if record.Error != "" {
-			out[prefix+"error"] = record.Error
+		if record.Cause != nil {
+			out[prefix+"error"] = record.ErrorKind
```
ksarangmath: should we just remove error now since it's the same thing as error kind?

